PR #142, opened 2023-01-11
docs: ADR for brew installation method

daniel-makerx:


docs/architecture-decisions/2023-01-11_brew_install.md
```diff
@@ -0,0 +1,86 @@
...
+This option uses a cask which does not have the maintenance overhead of a formula, and can be hosted in an Algorand Foundation repo to get around the fact AlgoKit does not meet the normal cask criteria.
+
+**Pros**
+* Algorand Foundation has control over this process and it can be fully automated
+* Lower maintenance cost as we do not need to support source builds of dependencies and it's consistent with how algokit cli is installed via Chocolatey and pipx
+* Fast install for all MacOS platforms
+* Fast build time on release
+
+**Cons**
+* Less discoverable install for end users `brew install --cask algorandfoundation/algokit`
```
daniel-makerx, line 72: I discovered --cask isn't actually required, I think its only necessary when there is both a formula and cask with the same name

```suggestion
* Less discoverable install for end users `brew install algorandfoundation/algokit`
```

PR #147, opened 2023-01-11
Added beaker productionisation review

robdmoore:


docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
```diff
@@ -0,0 +1,409 @@
...
+- **Renamings** - There are some clear parameters that make sense to rename for various reasons
+- **Key decorator improvements** - Refactor some of the Beaker decorators to fix some bugs and improve user experience
+- **Beaker state refactor** - Refactor of the Beaker state interfaces to improve user extensibility and significantly simplify the Beaker codebase to improve maintainability
+
+The recommended areas for future improvement are:
+
+- Typed client generation from app spec to improve deploy-time and run-time dev experience
+- `Tmpl` values in app spec so you can have type-safe deployment clients that substitute any template values reliably at contract deploy time
+- Refactor storage types (blob, reserved, etc.) to allow use of in-built Python types and operators (terser, more intuitive)
+- Box storage implementation improved to match local/global behaviour and also automatically delete itself on contract deletion
```
JasonWeathersby, line 53: Very interested to know how this would be implemented as you need box names even on a delete transaction.

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
```diff
@@ -0,0 +1,409 @@
...
+- Composable and stackable authorization and `@authorize` as a standalone decorator
+- PyTEAL typings to be improved to support types beyond `Expr` where a more explicit type can be specified (improves typing and extensibility)
```
JasonWeathersby, line 55: Can we get a list of the typings?

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
achidlow: With boxes being dynamic, yes it's not possible to completely handle this. However, given the limited Application storage, if you explicitly name a Box as a construct in Beaker, it would be good for it to be initialised the same was as Application storage and torn down automatically on deletion. In this way, for storage requirements known ahead of time (at least, ahead of create/update time), you can effectively have unlimited Application storage. This also raises the point of "migrations" which is something @barnjamin was keen on as well, I believe.

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
achidlow: So with PyTEAL's current architecture, `Expr` is effectively completely dynamically typed. The type is encoded into `Expr.type_of()`. If those types could be encoded in the Python type system instead (e.g. `UInt64Expr`, `BytesExpr`), it would make working with PyTEAL a lot easier and more intuitive.

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
bbroder-algo: Hello, is this to say you want to convert an empty box into additional key/value application storage space?

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
```diff
@@ -0,0 +1,409 @@
...
+- Support referencing an app/lsig via ID/address (deployed separately, potentially automatically as part reading a Directed Acyclic Graph (DAG) in application.json of application dependencies) or bytes (deployed inline, what was previously called precompile, noting this would be deploy-time substitution, not smart contract run-time substitute like `TemplateVariable`), this also may allow precompile to be deprecated (it's a very complex implementation for what we believe to be an advanced edge case)
```
bbroder-algo, line 56: I wonder if you've seen this open PR and if it might impact your thinking https://github.com/algorand/go-algorand/pull/4737

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
```diff
@@ -0,0 +1,409 @@
...
+
+   class MyApp(MyBaseApp):
+       pass
+   MyApp.counter.default = pyteal.Int(10)
+
+   class MyOtherApp(MyBaseApp):
+       pass
+
+   app1 = MyApp()
+   app2 = MyOtherApp()
```
bbroder-algo, line 95: I considered it and now my brain hurts.  It seems more like a python idiosyncrasy but it does illustrate your point.

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
```diff
@@ -0,0 +1,409 @@
...
+        output.set(app.state.counter),
+    )
+
+@app.external(authorize=Authorize.only(Global.creator_address()))
+def decrement(*, output: abi.Uint64):
+    """decrement the counter"""
+    return Seq(
+        app.state.counter.set(app.state.counter - Int(1)),
+        output.set(app.state.counter),
+    )
```
bbroder-algo, line 213: I see the flask-ness of the proposed changes with this example, thank you.

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
```diff
@@ -0,0 +1,409 @@
...
+
+The separation of compiled state outside of `Application` simplifies the design, and can be done mostly transparently to end-users.
+
+The separation of compiled state will also benefit future interoperability. It allows for more explicit decoupling of PyTEAL compilation (Beaker / PyTEAL transpilation -> TEAL) and deploment (TEAL -> compiled byte code -> Algorand network). Once `beaker.client` is split into a separate package, if the compiled state can be both generated from a beaker Application object _or_ loaded from disk (or similar), this means Beaker's ApplicationClient could be used in more situations, such as for a (say) tealish smart contract, or a C# smart contract, or a raw PyTEAL or TEAL smart contract, etc. This conforms better to the modularity principle in AlgoKit and also vice versa allows for a Beaker smart contract to be deployed by a TypeScript deployer, or C# deployer, etc.
+
+#### Before & After - user's perspective
+
+For most use cases, this should be a relatively small and probably imperceptible change.
+
+We believe there are two common usage scenarios that use the output of PyTEAL compilation currently:
```
bbroder-algo, line 239: You may also be interested in metadata such as the mapping of source code to line numbers.

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
```diff
@@ -0,0 +1,409 @@
...
+
+We suggest also potentially renaming `CompiledApplication.dump()`, perhaps to something along the lines of `serialize()`.
+
+The `compile()` call is actually a transpilation call (Beaker / PyTEAL transpilation -> TEAL), although it's called compile in PyTEAL so consideration should be made to either keep consistency with PyTEAL or use the more accurate `transpile()` (which also reduces confusion around the fact that you then have to call `compile` on algod to compile the TEAL to byte code before deployment).
+
+The exact details of what `CompiledApplication` will look like are TBD, but should be driven by the principles outlined in the "Why?" section above. Broadly, it stands to reason it would contain the approval and clear TEAL, the ABI spec and the app spec though at least.
+
+### (3) Renamings
+
+Renaming `version` parameter in `Application.__init__(version: int = pyteal.MAX_VERSION)` to (e.g.) `teal_version`, to be more explicit. Otherwise developers may be confused that it's the version of the specific smart contract. It may be desirable to allow `version` to continue to be specified for some time, but to raise a `DeprecationWarning`.
```
bbroder-algo, line 266: this one I'm not so sure about but I get the general idea.

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
achidlow: @bbroder-algo we should make it trivial for the user to treat them as such :)

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
achidlow: I had a brief look, and maybe I'm missing some subtler details, but would using macros not just be another way achieve the same outcome?

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
robdmoore: Cool, do you mind elaborating on your thoughts on this one?

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
robdmoore: This is a future item anyway, but good to know re: macros. Reminds me of C/C++ macros!

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
robdmoore: Good point. I've added a note about that. I don't think it has an impact on the design though.

docs/architecture-decisions/2023-01-11_beaker_productionisation_review.md
robdmoore: I assume this is about how it should be `avm_version`, not `teal_version`, which we agree with and is being implemented.

PR #151, opened 2023-01-13
fix: Fixing bugs in chocolatey install script

robdmoore:


robdmoore:
This fix is incorrect, closing.

PR #162, opened 2023-01-23
Smart contract deployment architecture

robdmoore:
This is hard to read in the diff, go here to see the easy to read version directly: https://github.com/algorandfoundation/algokit-cli/blob/smart-contract-deployment-architecture/docs/architecture-decisions/2023-01-12_smart-contract-deployment.md


docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
```diff
@@ -0,0 +1,375 @@
...
+- All provided smart contract templates are by default permanent when deployed to MainNet, but deletable elsewhere to facilitate an iterative development experience
+- Client code will include mechanisms to dynamically find deployed applications in LocalNet and TestNet environments to support delete/recreate flows and improve the developer experience
+- MainNet deployments will immediately (i.e. before any usage occurs) check that smart contracts are not upgradeable by the creator account by default (with an explicit opt-out option available for smart contracts that are meant to be upgradeable, which in turn will issue a warning to the developer to explain the implications)
+- MainNet deployments will immediately (i.e. before any usage occurs) check that smart contracts are not deletable by the creator account by default (with an explicit opt-out option available for smart contracts that are meant to be deletable, which in turn will issue a warning to the developer to explain the implications)
+
```
algoanne, line 207: +1 to the making contract immutable by default, so glad to see that. 

As for the various checks on that topic, what about checking that any inner app calls are calling apps that are immutable? Was that considered?

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
```diff
@@ -0,0 +1,375 @@
+# Smart Contract Deployment
+
+- **Status**: Draft
+- **Owner:** Rob Moore
+- **Deciders**: Anne Kenyon (Algorand Inc.), Alessandro Ferrari (Algorand Foundation), Fabrice Benhamouda (Algorand Inc.)
```
algoanne, line 5: Fabrice is at the Foundation

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
```diff
@@ -0,0 +1,375 @@
...
+
+- Add payloads to application creation transactions to allow them to be resolved by name for a known creator account using the indexer.
+- Allow application IDs to be resolved by name using environment variables for situations where a developer is using ephemeral creator accounts.
+- On-chain resolution will be left for developers to implement (likely by hardcoding or Global-state-based configurable application IDs).
+
+More sophisticated options could potentially by implemented in the future.
+
+### Automated vs manual deployments
+
+In order to support Continuous Delivery for smart contracts there is a need to support automated deployments of smart contracts via deployment pipeline.
```
algoanne, line 263: how important is this for algokit v1? Are we talking about CD while developing against Testnet or something? I am imagining that there are few use cases where people would actually have upgradeable smart contracts that they want to use CD to upgrade. I _can_ imagine such a case but I don't think it's a common one.

algoanne: I do of course see value in guiding a healthy deployment against Mainnet, I'm just not sure about the CD part.

algoanne (commented):
I agree with pretty much every decisions made in here, excepting the one comment. Thoughtful and well-explained as usual. 

I also think this covers a great set of features to offer for a v1. I was trying to think of missing pieces but I think this will be a good place to build from, and better to get it in the hands of builders and iterate.

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
fabrice102: Actually how do you check whether a contract if not upgradeable/deletable?
I'm guessing with beaker, this can be done relatively easily.
But in general, this may require formal analysis. I'm guessing that checking that inner app calls are to immutable apps can be done without formal analysis. This is maybe something that would require a static analysis tool.

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
fabrice102: One argument was that it is best if the same process is used to deploy on TestNet vs deploy on MainNet, so that there are fewer risks when moving into production.

fabrice102: I don't think it should be a CI/CD pipeline necessarily for MainNet, but more a support that allow easily deployment as when done locally on TestNet/Sandbox.
Also, I think this should not allow for MainNet upgrade of smart contract, because MainNet upgrade of smart contract should only be done with secure keys and algokit does not have a secure-enough way of storing keys for that purpose.

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
```diff
@@ -0,0 +1,375 @@
...
+
+If a contract is immutable, it does limit the ability to evolve the functionality over time based on user feedback and also discourages use of best practice software delivery techniques that encourage evolution and rapid deployment like Continuous Delivery.
+
+Another consideration in favour of immutability is smart contract auditing. If a smart contract is audited for security, but then the smart contract is upgradeable or is changed between when the audit occurred and when the contract was deployed to MainNet then the smart contract audit is somewhat invalidated and certainly any hashes that are provided of the smart contract code in an audit report will no longer match.
+
+There are techniques that can be used to allow for immutable smart contracts, but let them be evolved somewhat:
+
+- You could release a new version of a smart contract and include an ability for that smart contract to communicate to/from the existing smart contract to migrate state.
+- You could ensure that clients can dynamically find the latest smart contract so the application ID / address doesn't need to be hardcoded and the smart contract remains addressable.
```
fabrice102, line 193: In case of oracles, another point of consideration is that different applications may have different requirements if the oracle disappear (which can happen for example if the data provider lost their keys or stop operating the oracle). Some more-security/decentralized-focus applications may decide to stop working. Some may decide to change oracle.

So I think it may make sense to systematically have such oracle be non-upgradable.
But then put a proxy smart contract in front that redirects to the right oracle for the applications that would like some form of updatability.

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
```diff
@@ -0,0 +1,375 @@
...
+
+Lastly, it's worth noting that having fast development feedback loops during development and testing of smart contracts is likely a very useful feature to improve the development experience and speed. For this reason, allowing contracts to be upgraded or at the very least deleted (and then recreated) is likely very useful, but potentially when deployed to MainNet a switch could be made to disallow upgrading / deleting (as relevant).
+
+The goal of AlgoKit is to create a development experience that is productive and easy, but also one that is secure by default and helps developers fall into the pit of success. For that reason, and given the consequences of getting this wrong the suggested approach AlgoKit takes is:
+
+- All provided smart contract templates are by default immutable
+- An immutability automated test is included by default to ensure that smart contracts can't be upgraded by the contract creator (this would have to be deleted by a developer, who is then opting in to the consequences of that)
```
fabrice102, line 202: Would this be some analysis of the beaker code (which can thus be bypassed by users) or would this be done by prepending to the output TEAL code some code that directly prevent delete/upgrade at all?

I think that the first option is fine: if developers voluntarily bypass this protection using complex dirty tricks, they know what they are doing.

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
```diff
@@ -0,0 +1,375 @@
...
+- MainNet deployments will immediately (i.e. before any usage occurs) check that smart contracts are not upgradeable by the creator account by default (with an explicit opt-out option available for smart contracts that are meant to be upgradeable, which in turn will issue a warning to the developer to explain the implications)
+- MainNet deployments will immediately (i.e. before any usage occurs) check that smart contracts are not deletable by the creator account by default (with an explicit opt-out option available for smart contracts that are meant to be deletable, which in turn will issue a warning to the developer to explain the implications)
+
+### Mnemonic storage and retrieval
+
+When deploying and interacting with a smart contract, you need to have access to the private key of an account. This is a secret and must be handled with care, as exposing a private key can be disastrous, and while [rekeying](https://developer.algorand.org/docs/get-details/accounts/rekey/) is possible if it's not done fast enough you can still loose assets, be victim to malicious calls and experience a painful user experience going forward (wallet support for rekeyed accounts is limited).
```
fabrice102, line 210:
```suggestion
When deploying and interacting with a smart contract, you need to have access to the private key of an account. This is a secret and must be handled with care, as exposing a private key can be disastrous, and while [rekeying](https://developer.algorand.org/docs/get-details/accounts/rekey/) is possible if it's not done fast enough you can still lose assets, be victim to malicious calls and experience a painful user experience going forward (wallet support for rekeyed accounts is limited).
```

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
```diff
@@ -0,0 +1,375 @@
...
+It's important to consider whether the smart contract needs to be resolved on-chain or off-chain.
+
+Resolving on-chain is harder to achieve dynamically, but there are some patterns that can be used, e.g.:
+
+- Storing the application ID in (e.g. Global) state and providing a creator-only ABI method that can be called as part of deployment of the dependant contract to update the stored application ID.
+- [Lookup/registry contract](https://research.csiro.au/blockchainpatterns/general-patterns/contract-structural-patterns/contract-registry/) that returns the ID of a named contract and that lookup contract allows said ID to be updated.
+- [Proxy contract](https://blog.openzeppelin.com/proxy-patterns/) that mirrors the interface of the parent contract but delegates calls to an underlying contract whose ID can be updated.
+- [Other patterns](https://ethereum.org/en/developers/docs/smart-contracts/upgrading/#what-is-a-smart-contract-upgrade).
+
+Resolving off-chain can be done through a variety of ways, e.g.:
```
fabrice102, line 244: An alternative way may be to just have a note like

```
algokit/app:u<UUID>
```
(following [ARC2](https://arc.algorand.foundation/ARCs/arc-0002))
where <UUID> is a random UUID generated at initialization of the project.

However, this may not be a good option anymore because the `note-prefix` parameter is disabled by default in indexer.

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
```diff
@@ -0,0 +1,375 @@
...
+
+- The deployment pipeline needs to determine if the smart contract already exists or not in the network being deployed to
+- Depending on whether the contract is immutable and/or permanent and if it already exists or not in the network being deployed to, the pipeline needs to handle upgrade, deletion and/or creation of the contract (and this behaviour may need to be switched depending on the network being deployed to)
+- In order to protect integrity of MainNet contracts the deployment process should probably have some safeguards for MainNet that by default prevent destructive operations and require some kind of human opt-in to perform those operations
+- Exposing private key mnemonics for MainNet necessitates they are "hot wallets" which may be undesirable for privileged accounts responsible for high value smart contracts (e.g. smart contracts locking $m's or $b's), alternatively low privilege scenarios may benefit from (encrypted secret) mnemonic storage to improve developer experience and operational overhead
+- Supporting cold storage wallets for high privilege accounts, without resorting the manual deployments (which present their own risks) is tricky
```
fabrice102, line 271: I expect that deployment often just consists, at the end, of a group of transactions to send to the blockchain (maybe multiple groups).
One solution in that case is to output the unsigned transaction file, ask the user to take it and sign it and give it back, then send it.

But this would not be in a CI/CD pipeline, rather just a command line deployment process (`algokit deploy mainnet`).

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
robdmoore: It's a great question @fabrice102!

I think initially we can:

a) ask the developer to specify if the contract is upgradeable/permanent
b) include an automated test that calls the contract with a basic check

Longer term, formal analysis would be ideal, but that's more advanced than would be possible for v1.

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
robdmoore: > One argument was that it is best if the same process is used to deploy on TestNet vs deploy on MainNet, so that there are fewer risks when moving into production.

💯 

Exactly Fabrice, it's about providing confidence and rigour by ensuring that only code that has been through a defined process that incrementally increases confidence and reduces risk can be deployed to production (MainNet).

Is it essential for v1? Probably not, it's something I'm passionate about though :)

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
robdmoore: Makes sense.

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
robdmoore: My thought here was probably simpler - a test that executes the smart contract against localnet with the deployer account and tries to update the contract (expecting failure).

docs/architecture-decisions/2023-01-12_smart-contract-deployment.md
robdmoore: Noted (pun intended)

Issue #164, opened 2023-01-28
Gitpod installation - Doctor version parsing for docker compose


SilentRhetoric:
Running `algokit doctor` in a Gitpod where algokit has been installed (thanks @joe-p) throws the following error on docker combine version parsing:

Gitpod link:
[https://gitpod.io/#https://github.com/algorand-devrel/docker-algorand-gitpod/tree/algokit/](https://gitpod.io/#https://github.com/algorand-devrel/docker-algorand-gitpod/tree/algokit/
)

Terminal input & output:
```
gitpod /workspace/docker-algorand-gitpod (algokit) $ algokit doctor
timestamp: 2023-01-28T14:14:35+00:00
AlgoKit: 0.2.0
AlgoKit Python: 3.10.7 (main, Jan 28 2023, 13:07:35) [GCC 9.4.0] (location: /home/gitpod/.local/pipx/venvs/algokit)
OS: Linux-5.15.0-47-generic-x86_64-with-glibc2.31
docker: 20.10.17
docker compose: 2.10.0-gitpod.0
  Failed to parse version from: "2.10.0-gitpod.0"
  Error: invalid literal for int() with base 10: '0-gitpod'
  Unable to check against minimum version of 2.5.0
git: 2.37.3
python: 3.10.7 (location: /home/gitpod/.pyenv/shims/python)
python3: 3.10.7 (location: /home/gitpod/.pyenv/shims/python3)
pipx: 1.1.0
poetry: 1.3.1
node: 16.19.0
npm: 8.19.3
```

joe-p:
For checking min version the regex `\d+\.\d+\.\d+` should be used to parse version number from custom builds. There probably should be warning as well, but definitely not an error like this

joe-p:
The semver library might also be useful to handle other edge cases that we might not thinking of: https://pypi.org/project/semver/

robdmoore:
@daniel-makerx can you take a peek?

robdmoore:
The next Algokit release should fix this.

If you want to try it in the meantime you can `pipx install https://github.com/algorandfoundation/algokit-cli` to get the latest code on main

joe-p:
With #168 the version checking is fixed, however the output for the doctor command now omits the custom version string. For example:

```
$ docker-compose --version
Docker Compose version v2.10.0-gitpod.0

$ algokit doctor
timestamp: 2023-02-10T12:49:38+00:00
AlgoKit: 0.2.0
[...]

PR #173, opened 2023-02-10
fix(doctor): Ensuring full docker version information is visible in Doctor output to improve debugging

robdmoore:
Fixes comment in #164

### DIFF
--- a/src/algokit/core/doctor.py
+++ b/src/algokit/core/doctor.py
@@ -89,7 +89,10 @@ def check_dependency(
         )
     if minimum_version is not None:
         try:
-            version_ok = is_minimum_version(version_output, minimum_version)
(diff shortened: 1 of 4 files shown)


daniel-makerx (approved):

src/algokit/core/doctor.py
```diff
@@ -89,7 +89,10 @@ def check_dependency(
         )
     if minimum_version is not None:
         try:
-            version_ok = is_minimum_version(version_output, minimum_version)
+            match = re.search(r"^\d+\.\d+\.\d+", output)
```
achidlow, line 92: two things here:
- it was previously using `version_output` because that's the unaltered output
- raise to catch is a bit of an anti-pattern

PR #180, opened 2023-02-23
feat: add init --template-url-ref option to allow using a specific commit, tag or branch

daniel-makerx:


achidlow (changes requested):
Could we maybe add this as another CLI option, rather than potential breaking the parsing of `username@`?

achidlow (approved):

robdmoore (approved):

PR #188, opened 2023-02-25
Enhancement: adding template config file for easier updates of template options

barnjamin:
marked as draft until I actually test it

### DIFF
--- a/src/algokit/cli/init.py
+++ b/src/algokit/cli/init.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import re
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


src/algokit/cli/init.py
```diff
@@ -146,10 +149,10 @@ def init_command(
     template_name: str | None,
     template_url: str | None,
     template_url_ref: str | None,
-    unsafe_security_accept_template_url: bool,  # noqa: FBT001
+    unsafe_security_accept_template_url: bool,
```
barnjamin, line 152: i did not change this manually, some auto formatter thing?

PR #200, opened 2023-03-10
Try to launch VSCode after init if `.vscode` exists and `code` is in path

achidlow:


src/algokit/cli/init.py
```diff
@@ -238,8 +239,9 @@ def init_command(
         # if the URL looks like an HTTP URL (should be the case for blessed templates), be helpful
         # and print it out so the user can (depending on terminal) click it to open in browser
         logger.info(f"Your selected template comes from:\n➡️  {expanded_template_url.removesuffix('.git')}")
-    logger.info("As a suggestion, if you wanted to open the project in VS Code you could execute:")
-    logger.info(f"> cd {directory_name} && code .")
+    if shutil.which("code"):
+        logger.info("As a suggestion, if you wanted to open the project in VS Code you could execute:")
+        logger.info(f"> cd {directory_name} && code .")
```
robdmoore, line 244: Can we do an else with cd {directory name} and then open in your fav IDE?

robdmoore (approved):

src/algokit/cli/init.py
achidlow: I did consider that, but I'm not sure that adds much value? 
Actually, now that I consider it, you don't even need to cd there - we should just use `code {directory_name}`

PR #202, opened 2023-03-16
`goal` / `localnet console` exit code smarts

achidlow:
When executing goal/bash inside algod container only show localnet hint if it looks like the container doesn't exist or isn't running.

Also update ruff (so that we have pattern matching support).

### DIFF
--- a/src/algokit/core/sandbox.py
+++ b/src/algokit/core/sandbox.py
@@ -94,8 +94,10 @@ class ComposeSandbox:
             bad_return_code_error_message="Failed to get logs, are the containers running?",
         )
 
-    def ps(self) -> list[dict[str, Any]]:
-        run_results = self._run_compose_command("ps --format json", stdout_log_level=logging.DEBUG)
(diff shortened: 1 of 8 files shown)


src/algokit/core/sandbox.py
```diff
@@ -94,8 +94,8 @@ def logs(self, *, follow: bool = False, no_color: bool = False, tail: str | None
             bad_return_code_error_message="Failed to get logs, are the containers running?",
         )
 
-    def ps(self) -> list[dict[str, Any]]:
-        run_results = self._run_compose_command("ps --format json", stdout_log_level=logging.DEBUG)
+    def ps(self, servie_name: str | None = None) -> list[dict[str, Any]]:
```
daniel-makerx, line 97: Is this meant to be service_name?

achidlow: ...maybe

daniel-makerx (approved):

PR #215, opened 2023-03-27
feat: prompt for template first

achidlow:
this allows the user to see the templates available and their description without having to supply any other parameters first

some minor refactoring also to speed this up on first run

### DIFF
--- a/src/algokit/cli/init.py
+++ b/src/algokit/cli/init.py
@@ -94,7 +94,7 @@ def validate_dir_name(context: click.Context, param: click.Parameter, value: str
     "-t",
     type=click.Choice(list(_get_blessed_templates())),
     default=None,
-    help="Name of an official template to use.",
+    help="Name of an official template to use. To see a list of descriptions, run this command with no arguments.",
 )
 @click.option(
     "--template-url",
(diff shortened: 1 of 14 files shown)


src/algokit/cli/init.py
```diff
@@ -165,55 +165,38 @@ def init_command(
...
-
     logger.debug(f"Attempting to initialise project in {project_path} from template {template}")
 
-    copier_worker = copier.run_copy(
+    # copier is lazy imported for two reasons
+    # 1. it is slow to import on first execution after installing
+    # 2. the import fails if git is not installed (which we check above)
+    # TODO: copier is typed, need to figure out how to force mypy to accept that or submit a PR
+    #       to their repo to include py.typed file
+    from copier.main import run_copy  # type: ignore[import]
```
achidlow, line 197: I moved this all the way down here, because this is a slow point anyway due to network fetch

src/algokit/cli/init.py
```diff
@@ -165,55 +165,38 @@ def init_command(
...
-    # copier is lazy imported for two reasons
-    # 1. it is slow to import on first execution after installing
-    # 2. the import fails if git is not installed
 
-    try:
-        # TODO: copier is typed, need to figure out how to force mypy to accept that or submit a PR
-        #       to their repo to include py.typed file
-        import copier  # type: ignore[import]
-    except ImportError as ex:
+    if not shutil.which("git"):
```
achidlow, line 169: I've checked the implementation of `which` that `plumbum` uses (ie what copier uses under the hood). It's a very old library, dating back to at least Python 2.5, so they don't use the built-in `shutil.which` that was added in 3.3, but the implementations are essentially equivalent. So if this check passes, we can be confident that copier will import successfully.

daniel-makerx (approved):

PR #216, opened 2023-03-28
Improved getting started

achidlow:
- point out README file in template (if exists) or open this file in VSCode
- let the user know when template execution is starting and when it's finished, to help provide context about where they are in the process
- improve debug logging in init
- simplify `platform` module mocking in tests


src/algokit/cli/init.py
```diff
@@ -181,32 +181,35 @@ def init_command(
...
 
     project_path = _get_project_path(directory_name)
+    logger.debug(f"project path = {project_path}")
     directory_name = project_path.name
     # provide the directory name as an answer to the template, if not explicitly overridden by user
     answers_dict.setdefault("project_name", directory_name)
 
-    logger.debug(f"Attempting to initialise project in {project_path} from template {template}")
-
+    logger.info("Starting template execution...")
```
robdmoore, line 192:
```suggestion
    logger.info("Starting template initialisation. Initialising in {project_path} from template {template}.")
```

src/algokit/cli/init.py
```diff
@@ -181,32 +181,35 @@ def init_command(
...
         vcs_ref=template.commit,
-    )
+    ) as copier_worker:
+        expanded_template_url = copier_worker.template.url_expanded
+        logger.debug(f"final clone URL = {expanded_template_url}")
+        copier_worker.run_copy()
 
-    expanded_template_url = copier_worker.template.url_expanded
-    logger.debug(f"Template initialisation complete, final clone URL = {expanded_template_url}")
+    logger.info("Template execution complete!")
```
robdmoore, line 212:
```suggestion
    logger.info("Template initialisation complete.")
```

robdmoore (approved):

src/algokit/cli/init.py
achidlow: So I went with execution over initialization because we use initialization later on, didn't want to confuse things.
Also those values are logged immediately as debug, and the user just provided them, so to keep the output clean and easier to digest I left them out of this message here

PR #238, opened 2023-04-04
Add Gitpod check for localnet explore

joe-p:
If a user is running algokit in gitpod (which we use in bootcamps), running `algokit localnet explore` will fail because dappflow won't be able to connect to `localhost`. This is a minor change that checks whether algokit is running in gitpod and modifies the localnet connection information accordingly. 

End result is `algokit localnet explore` working the same on local machine and on gitpod


daniel-makerx:
Thanks @joe-p, happy to merge this PR. Just needs the approval tests updated after changing the URL, could you rerun the tests and commit the updated approvals?

PR #240, opened 2023-04-06
mention the additional repositories

abebeos:
Make it clear that algokit-cli integrates multiple additional repositories.


robdmoore:
Thanks, I've incorporated the search, but in a more relevant spot. Worth noting the individual repositories were already linked to in the waht is algokit section.

abebeos:
> but in a more relevant spot.

This is not "a more relevant spot", it's the wrong spot (i don't even see it when opening the document).

You, as a core-dev, should avoid documenting you own stuff. 

Simply because you fail to perceive the problems that a newcomer has with your documentation.

abebeos:
@abebeos , you should really focus on implementation, and let contributors contribute. I abort at this point my attempt to streamline the documentation for newcomers. Takes anyways far far too long.

PR #252, opened 2023-04-21
mention Hyper-V Docker backend alternative

### LINKED ISSUE #232: Mention optional "Hyper-V" for Docker on Windows
> Some people have reported problems running WSL 2 on a small proportion of Windows environments (to get the latest Docker experience)
[source](https://github.com/algorandfoundation/algokit-cli/blob/main/docs/architecture-decisions/2022-11-14_sandbox-approach.md#option-1---pre-built-dockerhub-images)

It should be mentioned within the relevant documentation that users can simply switch to Hyper-V if they have problems with WSL2.

---
Personally, after many troubles with WSL2 around 2 years ago(?), I switched back to a Hyper-V backend for Docker. Had nearly never problems again (e.g excessive memory consumptions of WSL2, in some scenarios terribly slow filesystem-access-speed, incompatibilities and strangeness's)

abebeos:
Fixes #232

### DIFF
--- a/docs/features/localnet.md
+++ b/docs/features/localnet.md
@@ -14,6 +14,8 @@ You can install Docker by following the [official installation instructions](htt
 
 If you are on Windows then you will need WSL 2 installed first, for which you can find the [official installation instructions](https://learn.microsoft.com/en-us/windows/wsl/install). If you are using Windows 10 then ensure you are on the latest version to reduce likelihood of installation problems.
 
+If you prefer so, the Windows 10 Pro supported Hyper-V Docker backend can be used instead of the WLS2 backend.
+
 ## Known issues
 
 The AlgoKit LocalNet is built with 30,000 participation keys generated and after 30,000 rounds is reached it will no longer be able to add rounds. At this point you can simply reset the LocalNet to continue development. Participation keys are slow to generate hence why they are pre-generated to improve experience.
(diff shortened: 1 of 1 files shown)


docs/features/localnet.md
```diff
@@ -14,6 +14,8 @@ You can install Docker by following the [official installation instructions](htt
 
 If you are on Windows then you will need WSL 2 installed first, for which you can find the [official installation instructions](https://learn.microsoft.com/en-us/windows/wsl/install). If you are using Windows 10 then ensure you are on the latest version to reduce likelihood of installation problems.
 
+If you prefer so, the Windows 10 Pro supported Hyper-V Docker backend can be used instead of the WLS2 backend.
```
robdmoore, line 17:
```suggestion
Alternatively, the Windows 10/11 Pro+ supported [Hyper-V backend](https://docs.docker.com/desktop/install/windows-install/) for Docker can be used instead of the WSL 2 backend.
```

abebeos:
After looking at the end-result:

I believe it would be enough to just point to the docker documentation:

> AlgoKit LocalNet relies on Docker and Docker Compose being present and running on your system.

> You can install Docker by following the [official installation instructions](https://docs.docker.com/get-docker/).

It contains everything needed, including WSL2, Hyper-V (and docker-compose).

abebeos:
see follow-up #254

robdmoore:
I prefer this version. We've had lots of feedback via the dev rels that this is an area people struggle with so I'm happy to be more explicit and set expectations about what they need to look for even if we are doubling up a bit.

PR #273, opened 2023-06-01
feat: Adding minimum required version for algokit.

negar-abbasi:


src/algokit/core/bootstrap.py
```diff
@@ -9,6 +9,10 @@
 
 from algokit.core import proc, questionary_extensions
 
+import toml
+from algokit.core.version_prompt import get_latest_version_or_cached
```
daniel-makerx, line 13: this function returns the latest version available on GitHub, for this check we want to check the current *executing* version, to do that use this 
`from algokit.core.conf import get_current_package_version`

src/algokit/core/bootstrap.py
```diff
@@ -218,3 +221,24 @@ def _get_base_python_path() -> str | None:
                 return str(candidate_path)
     # give up, we tried...
     return this_python
+
+
+def version_check(cwd: Path) -> None:
+    """
+    Checks the current version of AlgoKit against the minimum required version specified in algokit.toml.
```
daniel-makerx, line 228:
```suggestion
    Checks the current version of AlgoKit against the minimum required version specified in .algokit.toml.
```

PR #287, opened 2023-06-15
Update docker version command

joe-p:
When using a custom build of docker, `docker compose version` will only return the commit hash and not the release version.

```
$ docker compose version
Docker Compose version dde0213

$ docker compose version --format json
{"version":"dde0213"}
```

For the version parsing to work, `docker --version` must be used:

```
$ docker --version
Docker version 23.0.3, build 3e7cbfd
```

### DIFF
--- a/src/algokit/core/sandbox.py
+++ b/src/algokit/core/sandbox.py
@@ -284,4 +284,4 @@ def fetch_indexer_status_data(service_info: dict[str, Any]) -> dict[str, Any]:
         return {"Status": "Error"}
 
 
-DOCKER_COMPOSE_VERSION_COMMAND = ["docker", "compose", "version", "--format", "json"]
+DOCKER_COMPOSE_VERSION_COMMAND = ["docker", "--version"]


src/algokit/core/sandbox.py
```diff
@@ -284,4 +284,4 @@ def fetch_indexer_status_data(service_info: dict[str, Any]) -> dict[str, Any]:
         return {"Status": "Error"}
 
 
-DOCKER_COMPOSE_VERSION_COMMAND = ["docker", "compose", "version", "--format", "json"]
+DOCKER_COMPOSE_VERSION_COMMAND = ["docker", "--version"]
```
robdmoore, line 287: This is the docker version rather than docker compose, if we used this we'd need to change the version we look for (we look for compose 2.5+ from memory because that was the point it was stable with bug patches for things users of algokit would likely come across.

robdmoore:
Looping back here, what should we do @joe-p? When does a custom docker build get used?

PR #288, opened 2023-06-15
PyInstaller

joe-p:
Adds PyInstaller to the dependencies and a poe task for building the binary. This allows us to distribute a binary without needing users to install python and pipx first, which is particularly useful for devs not using Python in their development stack but still want to use algokit. 

I still need to do some more testing, but so far things seem to be working on Mac (ARM) and Ubuntu.


.github/workflows/check-python.yaml
```diff
@@ -48,3 +48,6 @@ jobs:
 
       - name: Check docs are up to date
         run: poetry run poe docs && [ $(git status --porcelain docs/ | wc -l) -eq "0" ]
+
+      - name: Check PyInstaller compilation
+        run: poetry run poe build
```
robdmoore, line 53: We probably want this in the build one rather than the check one. And then probably should add it as an artifact?

pyproject.toml
```diff
@@ -7,7 +7,7 @@ license = "MIT"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = "^3.10"
+python = "<3.12,>=3.10"
```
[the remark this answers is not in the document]
joe-p: I fixed this in [1b587e1](https://github.com/algorandfoundation/algokit-cli/pull/288/commits/1b587e1773b171644f083ebe28fed448f2dafa87). <3.12 is only necessary for PyInstaller itself so I moved it to dev deps and specified the python version explicitly for PyInstaller.

.github/workflows/check-python.yaml
joe-p: Just wanted to have it here as an initial smoke test but yeah I agree, before we merge we should make an action for automatically building it and including zips in the release.

joe-p:
I'd like to bring this into consideration again. Main thing missing that should be here before merge is integrating into workflows. 

Also, I did some benchmarking for the PyInstaller folder build vs Nuitka vs pipx install. They all seem to be pretty close with eachother so we don't loose much going from pipx > pyinstaller. PyInstaller is marginally faster than Nuitka, but since PyInstaller is more popular probably makes sense to stick with it. 

Note that this is for the directory distribution. It *can* be compiled to a single binary, but performance is much worse. This is totally fine for `brew`/`choco` since they can just put the files in PATH and call it a day. I think eventually we'd want an msi installer for windows for those that don't have `choco`. 

```
❯ hyperfine --warmup 3 -i "algokit doctor" "./pyinstaller-dist/algokit/algokit doctor" "nuitka-dist/algokit.bin doctor"
Benchmark 1: algokit doctor
  Time (mean ± σ):      1.220 s ±  0.060 s    [User: 0.645 s, System: 0.211 s]
  Range (min … max):    1.160 s …  1.353 s    10 runs
 
Benchmark 2: ./pyinstaller-dist/algokit/algokit doctor
  Time (mean ± σ):      1.251 s ±  0.032 s    [User: 0.696 s, System: 0.198 s]
  Range (min … max):    1.211 s …  1.299 s    10 runs
 
Benchmark 3: nuitka-dist/algokit.bin doctor
  Time (mean ± σ):      1.268 s ±  0.033 s    [User: 0.726 s, System: 0.186 s]
  Range (min … max):    1.232 s …  1.339 s    10 runs
 
Summary
  algokit doctor ran
    1.02 ± 0.06 times faster than ./pyinstaller-dist/algokit/algokit doctor
    1.04 ± 0.06 times faster than nuitka-dist/algokit.bin doctor
```

aorumbayev:
@joe-p thanks again for all the work on this PR so far! Just fyi, i moved your latest and merged latest from main into https://github.com/algorandfoundation/algokit-cli/pull/382. @negar-abbasi will be spiking an ADR to asses pyinstaller and few other alternatives - there are still some limitations with pyinstaller+gh actions when it comes to apple silicon builds and linux distros with glib dependency - will be assessing how to address those as part of ADR. As for this PR we might revisit and close it once we get to implementation (cc @neilcampbell )

PR #295, opened 2023-06-28
feat: deploy command

aorumbayev:
# Proposed changes

- Introducing deploy command with the following behaviour:
- - By default grabs deploy configurations from hardcoded settings that rely on algonode for testnet/betanet/mainnet and localnet values for sandbox 
- - By default assumes that network specific deployment command is specified in algokit.toml file otherwise expects it to be passed via the argument
- - If deployment is being performed to mainnet it will explicitly warn user whether he wants to continue otherwise user can pass --prod flag to skip the warning prompt
- - Deployer mnemonic is required to be passed explicitly via masked input interactively or directly as an argument (useful in CI), docs are to be improved to explicitly make it clear that dispenser is optional while deployer is required if deployment is being done to non localnet networks 
- - As discussed and agreed with @achidlow @Loedn and @daniel-makerx , we are going ahead with support for .env.{any_network_name_or_default_networks} approach. That way user still has an option to pick our default localnet|betanet|testnet|mainnet algonode configs OR he defines {any_network_name_or_default_networks} in algokit.toml (or passed directly via flag) AND has a .env.[any_network_name_or_default_networks] to define network configurations

# Examples

algokit.toml
```
[algokit]
min_version = "version_goes_her"

[deploy.localnet]
command = "command_goes_here"

[deploy.testnet]
command = "command_goes here"

OR user can have any custom name basically
[deploy.{any_network_name_or_default_networks}]
command = "command_goes_here"
and so on...
```

.env.[any_network_name_or_default_networks] file with expectations (same content as current .env content on beaker templates)
```
# Sandbox
ALGOD_TOKEN=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
ALGOD_SERVER=http://localhost
ALGOD_PORT=4001
INDEXER_TOKEN=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
INDEXER_SERVER=http://localhost
INDEXER_PORT=8980
```

User can also supply a single .env file without network name prefix, in this case algokit will load that if no env network prefixed file is available. This is useful for cases when deployments are always done to a single network.

Example commands:

```
algokit deploy {any_network_name_or_default_networks} 
algokit deploy {any_network_name_or_default_networks} --ci --prod 
algokit deploy localnet --custom-deploy command "echo HelloWorld" 
```

# TODO 
- [x] Unit/Snapshot tests
- [x] Documentation
- [x] Propagating changes to beaker prod template (env file has to be updated as well as algokit toml file


aorumbayev:
@inaie-makerx @negar-abbasi tagging you for extra visibility so you are aware in case it overlaps with your items in any way. This will also involve some breaking changes on the beaker template mainly .env files and other configs would need to be placed at the root

src/algokit/core/typed_client_generation.py
```diff
@@ -27,8 +27,8 @@ class ClientGenerator(abc.ABC):
     language: ClassVar[str]
     extension: ClassVar[str]
 
-    _by_language: dict[str, type["ClientGenerator"]] = {}
-    _by_extension: dict[str, type["ClientGenerator"]] = {}
+    _by_language: dict[str, type["ClientGenerator"]] = {}  # noqa: RUF012
+    _by_extension: dict[str, type["ClientGenerator"]] = {}  # noqa: RUF012
```
achidlow, line 31: these shouldn't be suppressed, we should add the `ClassVar` annotation like the rule suggests here.

src/algokit/core/log_handlers.py
```diff
@@ -27,7 +27,7 @@ class ClickHandler(logging.Handler):
     but they're kind of intertwined for our use case of actually displaying things to the user.
     """
 
-    styles: dict[str, dict[str, Any]] = {
+    styles: dict[str, dict[str, Any]] = {  # noqa: RUF012
```
achidlow, line 30: this one can also be `ClassVar` rather than suppressing.

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,66 @@
...
+    required=False,
+    prompt=True,
+    default=lambda: HiddenSecret(os.environ.get(DISPENSER_KEY, "")),
+    hide_input=True,
+)
+@click.option(
+    "--custom_deploy_command",
+    type=str,
+    default=None,
+    help="Custom deploy command. If not provided, will load the deploy command from .algokit.toml file in the current directory.",  # noqa: E501
```
achidlow, line 43: don't suppress this one, split the strings by bracketing like:
```
(
"long"
"string"
)
```

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,66 @@
...
+)
+def deploy_command(network: str, deployer_mnemonic: str, dispenser_mnemonic: str, custom_deploy_command: str) -> None:
+    """Deploy smart contracts from AlgoKit compliant repository.
+    Will load the deploy command from .algokit.toml file in the current directory.
+    Will load default configuration for network using AlgoNode. To override the default configuration,
+    create a .env file in the current directory and set the required environment variables prefixed with the network name (e.g. LOCALNET_ALGOD_TOKEN and etc)
+    """  # noqa: E501
```
achidlow, line 50: also don't suppress here, definitely no need for that in a docstring

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,66 @@
...
+    # Load deploy config
+    project_dir = Path.cwd()
+    deploy_config = load_deploy_config(network, project_dir)
+
+    # Inject deploy config with mnemonic
+    deploy_config[DEPLOYER_KEY] = deployer_mnemonic
+    deploy_config[DISPENSER_KEY] = dispenser_mnemonic
+
+    # Load deploy command
+    deploy_command = custom_deploy_command or load_deploy_command(network_name=network, project_dir=Path.cwd())
```
achidlow, line 61: could just be `project_dir=project_dir`?

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+DEPLOYER_KEY = "DEPLOYER_MNEMONIC"
+DISPENSER_KEY = "DISPENSER_MNEMONIC"
+
+
+@click.command("deploy")
+@click.argument("network", type=click.Choice(list(DEPLOY_CONFIGS)), default="localnet", required=True)
+@click.option(
+    "--deployer_mnemonic",
```
achidlow, line 27: command line options should be dashed not snake cased

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,66 @@
...
+    required=True,
+    default=lambda: HiddenSecret(os.environ.get(DEPLOYER_KEY, "")),
```
achidlow, line 29: can you use click support for env-vars here? https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # Fill missing keys with empty strings
+        for required_key in REQUIRED_KEYS:
+            if required_key not in network_config:
+                logger.warning(f"Warning: {required_key} is missing, filling with an empty string.")
+                network_config[required_key] = ""
+
+        return network_config
+    else:
+        logger.info(f"No .env file found, loading default configuration for {network}.")
```
achidlow, line 107: I recommend inverting this condition.
In general, shortest path first is easier for a human to read - limited working memory slots means on a straight read through, the original context of this condition can be lost.

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,157 @@
...
+    :return: Deploy command.
+    """
+
+    # Load and parse the TOML configuration file
+    config_path = project_dir / ALGOKIT_CONFIG
+    try:
+        config_content = config_path.read_text("utf-8")
+    except FileNotFoundError:
+        logger.debug(f"No {ALGOKIT_CONFIG} file found in the project directory.")
+        raise click.ClickException(f"No {ALGOKIT_CONFIG} file found in the project directory.") from None
```
achidlow, line 125: why not raise from the original exception instead of `None` here?

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,157 @@
...
+
+    try:
+        config = tomllib.loads(config_content)
+    except FileNotFoundError:
```
achidlow, line 129: this doesn't seem like the right exception to check for here, the file is already loaded and read?

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    Parse the .env file if it exists and load configurations.
+    :param network: Network name.
+    :param project_dir: Project directory path.
+    :return: A dictionary containing the configuration.
+    """
+
+    env_file_path = project_dir / ".env"
+    if env_file_path.exists():
+        config = configparser.ConfigParser()
```
achidlow, line 86: we could use https://github.com/theskumar/python-dotenv#load-configuration-without-altering-the-environment instead here?

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,157 @@
...
+DEPLOY_CONFIGS = {
+    "localnet": _generate_config(
+        algod_token="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+        algod_server="http://localhost",
+        algod_port="4001",
+        indexer_token="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+        indexer_server="http://localhost",
+        indexer_port="8980",
+    ),
+    "testnet": _generate_config(
```
achidlow, line 61: this is duplicating some stuff from `cli/explore.py`, maybe we should refactor here?

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,157 @@
...
+def load_deploy_command(network_name: str, project_dir: Path) -> str:
+    """
+    Load the deploy command for the given network from .algokit.toml file.
+    :param network_name: Network name.
+    :param project_dir: Project directory path.
+    :return: Deploy command.
+    """
+
+    # Load and parse the TOML configuration file
+    config_path = project_dir / ALGOKIT_CONFIG
```
achidlow, line 120: it might be better to have a AlgoKit config load function in `core/conf.py` that both here and `core/bootstrap.py` can reference?

src/algokit/core/typed_client_generation.py
aorumbayev: Done

src/algokit/cli/deploy.py
aorumbayev: Done

src/algokit/cli/deploy.py
aorumbayev: Done

src/algokit/cli/deploy.py
aorumbayev: I've refactored it but using a different approach, please check the latest code @achidlow

src/algokit/core/deploy.py
aorumbayev: Done

src/algokit/core/deploy.py
aorumbayev: added new dep to poetry - now relies on dotent package

src/algokit/core/deploy.py
aorumbayev: moved them out to new constants.py that also aggregates a bunch of other env vars reused across core

src/algokit/core/deploy.py
aorumbayev: loading is not in conf.py , please check the latest code @achidlow

aorumbayev:
@achidlow @daniel-makerx pushed next batch of tweaks, still need some more unit tests to be added but the approach is now refined as agreed with Adam

src/algokit/cli/deploy.py
aorumbayev: Resolved

src/algokit/cli/deploy.py
aorumbayev: Resolved

src/algokit/core/log_handlers.py
aorumbayev: Resolved

src/algokit/core/deploy.py
aorumbayev: Resolved

src/algokit/core/deploy.py
aorumbayev: Resolved

docs/features/deploy.md
```diff
@@ -0,0 +1,117 @@
...
+For instance, if you're aiming to deploy to a testnet, your `.algokit.toml` file should contain:
+
+```toml
+[deploy.testnet]
+command = "poetry run python -m smart_contracts deploy"
+```
+
+### Environment Files
+
+For each network you're deploying to, you need to have a corresponding `.env.[network_name]` file. This file should contain environment variables specific to that network.
```
robdmoore, line 51: This says you need the file, it’s optional right?

aorumbayev: @robdmoore see the `note` field right below it explicitly saying its optional

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,108 @@
...
+    execute_deploy_command,
+    get_genesis_network_name,
+    load_deploy_command,
+    load_deploy_config,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def extract_mnemonics(skip_mnemonics_prompts: bool, network: str) -> tuple[str | None, str | None]:  # noqa: FBT001
```
achidlow, line 18: in general, don't suppress linting or mypy errors without a good reason - usually that reason is because a call signature or whatever is dictated by external code. here it's not, you can just make `skip_mnemonics_prompts` kwargs-only.

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    Extract environment variables, prompt user if needed.
+
+    :param skip_mnemonics_prompts: A boolean indicating if user prompt should be skipped.
+    :param network: The name of the network ('localnet', 'testnet', 'mainnet', etc.)
+    :return: A tuple containing deployer_mnemonic and dispenser_mnemonic.
+    """
+    deployer_mnemonic = os.environ.get(DEPLOYER_KEY, "")
+    dispenser_mnemonic = os.environ.get(DISPENSER_KEY, None)
```
achidlow, line 27: why is one of these defaulting to `""` and the other to `None`?
also, you can just use `os.getenv`

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,108 @@
...
+
+    if is_deployer_mnemonic_empty and is_not_localnet:
+        deployer_mnemonic = click.prompt("deployer-mnemonic", hide_input=True)
+
+    if (
+        is_dispenser_mnemonic_empty
+        and is_not_localnet
+        and click.confirm("Do you want to use a dispenser account?", default=False)
+    ):
+        dispenser_mnemonic = click.prompt("dispenser-mnemonic", hide_input=True)
```
achidlow, line 41: this whole structure using booleans is very confusing to me.

wouldn't it be simpler to structure with if statements?

like:
```
    if network != LOCALNET and not skip_mnemonics_prompts:
        if not deployer_mnemonic:
            deployer_mnemonic = click.prompt("deployer-mnemonic", hide_input=True)

        if not dispenser_mnemonic:
            use_dispenser = click.confirm("Do you want to use a dispenser account?", default=False)
            if use_dispenser:
                dispenser_mnemonic = click.prompt("dispenser-mnemonic", hide_input=True)
```

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,108 @@
...
+
+    return deployer_mnemonic, dispenser_mnemonic
+
+
+@click.command("deploy")
+@click.argument(
+    "network", type=str, default=LOCALNET, required=True, callback=(lambda _ctx, _param, value: value.lower())
```
achidlow, line 48: I had to look up what `callback` does to be sure - wouldn't it be simpler to have the `.lower()` call inside the command function?
also `type=str` is the default
also, what does it mean for an argument to have a default but also be required?

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,108 @@
...
+    "is_production_environment",
+    "--prod",
+    is_flag=True,
+    default=False,
+    help="Skip warning prompt for deployments to a mainnet.",
+)
+def deploy_command(
+    network: str,
+    custom_deploy_command: str,
+    skip_mnemonics_prompts: bool,  # noqa: FBT001
```
achidlow, line 74: you can make all args kwarg-only, click will handle it

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,108 @@
...
+    is_production_environment: bool,  # noqa: FBT001
+) -> None:
+    """Deploy smart contracts from AlgoKit compliant repository."""
+    logger.info(f"Starting deployment process on {network} network...")
+
+    project_dir = Path.cwd()
+    logger.info(f"Project directory: {project_dir}")
+
+    deploy_command = custom_deploy_command or load_deploy_command(network_name=network, project_dir=project_dir)
```
achidlow, line 83: so this only looking in `cwd` means that to execute this command, you must be in the root directory? so if I'm in a subdirectory in the project, it won't work?

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,108 @@
...
+    deployer_mnemonic, dispenser_mnemonic = extract_mnemonics(skip_mnemonics_prompts, network)
+
+    if not deployer_mnemonic and network != LOCALNET:
+        raise click.ClickException("Error: Deployer Mnemonic must be provided via env var or via cli input.")
+
+    deploy_config = load_deploy_config(network, project_dir)
+    logger.info("Loaded deployment configuration.")
+
+    if not is_production_environment:
+        network_name = network if network in DEFAULT_NETWORKS else get_genesis_network_name(deploy_config) or network
```
achidlow, line 95: I think this is complex enough logic to expand rather than use a ternary expression

src/algokit/core/bootstrap.py
```diff
@@ -7,29 +8,30 @@
 
 import click
 
+from algokit.core.constants import ALGOKIT_CONFIG
+
 if sys.version_info >= (3, 11):
-    import tomllib
+    pass
```
achidlow, line 14: this looks left over

src/algokit/core/bootstrap.py
```diff
@@ -1,3 +1,4 @@
+import glob
```
achidlow, line 1: Pathlib has build in globbing support

src/algokit/core/constants.py
```diff
@@ -0,0 +1,89 @@
+# mypy: disable-error-code="typeddict-unknown-key"
```
achidlow, line 1: hmmm what's the rational for disabling this mypy check? can we structure the code so that it passes without it instead?

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+import click
+import httpx
+from dotenv import dotenv_values
+
+from algokit.core import proc
+from algokit.core.conf import get_algokit_config
+from algokit.core.constants import ALGOKIT_CONFIG, ALGORAND_NETWORKS, AlgorandNetworkConfiguration
+
+if sys.version_info >= (3, 11):
+    pass
```
achidlow, line 20: this looks left over

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+    :return: Network name.
+    """
+
+    algod_server = deploy_config.get("ALGOD_SERVER")
+
+    if not algod_server:
+        raise click.ClickException("Missing ALGOD_SERVER in deploy configuration.")
+
+    try:
+        genesis_response = httpx.get(f"{algod_server}/genesis")
```
achidlow, line 40: what if there is a custom port?

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+        genesis_response.raise_for_status()
+        return genesis_response.json()["network"] if genesis_response.status_code == 200 else None  # noqa: PLR2004
```
achidlow, line 42: you can avoid this linting error with using `httpx.codes.OK` instead of `200`

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+    except httpx.HTTPError:
+        logger.warning(f"Failed to load network name from {algod_server} due to HTTP error.", exc_info=True)
```
achidlow, line 44: what about other errors? such as json key being missing?

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+            logger.info(f"No .env file found at {env_file_path}, loading default configuration for {network}.")
+            return dict(ALGORAND_NETWORKS.get(network, {}))
+
+        raise click.ClickException(f"No .env file found at {env_file_path}")
+
+    # Extract network configuration
+    network_config = dotenv_values(env_file_path, verbose=True, interpolate=True)
+
+    # Fill missing keys with empty strings
+    for required_key in list(AlgorandNetworkConfiguration.__annotations__.keys()):
```
achidlow, line 68: `for key in list(some_dict.keys())` is doubly redundant. iteration over keys is the default, and you don't need to convert to `list` to iterate (unless maybe you're mutating the underlying collection or something like that)

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+        if required_key not in network_config:
```
achidlow, line 69: is `INDEXER_SERVER` always required? doesn't that depend on the command?

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+
+def load_deploy_config(network: str, project_dir: Path) -> dict:
+    """
+    Load the deploy configuration for the given network.
+    :param network: Network name.
+    :param project_dir: Project directory path.
+    :return: Deploy configuration.
+    """
+
+    env_file_path = project_dir / f".env.{network}"
```
achidlow, line 56: we should be loading from `.env` also if it exists - `.env.{foo}` should take precedence though.
it might be easier here to:
1) copy `os.environ`
2) if network is a known name, update the dict with the defaults
3) load in from .env
4) load in from .env.foo if it it exists. if it doesn't exist and foo isn't a known network name, maybe that's an error?

achidlow: also, this will suffer from the same problem of requiring the user to be in the project root to run deploy, maybe it should walk up to discover env file(s) (and maybe also this should be done for algokit.toml - not sure what to do if that doesn't exist though)

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+    if not env_file_path.exists():
+        if network in ALGORAND_NETWORKS:
+            logger.info(f"No .env file found at {env_file_path}, loading default configuration for {network}.")
+            return dict(ALGORAND_NETWORKS.get(network, {}))
```
achidlow, line 60: `.get` is redundant if you've already done an `in` check

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    # Extract the deploy command for the given network
+    try:
+        return str(config["deploy"][network_name]["command"])
+    except KeyError:
+        raise click.ClickException(f"Deploy command is not specified in '{ALGOKIT_CONFIG}' file.") from None
+
+
+def execute_deploy_command(command: str, deploy_config: dict, project_dir: Path | None) -> None:
```
achidlow, line 100: project_dir is always passed so we don't need to make `None`-able

src/algokit/core/deploy.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Parse the deploy command
+    try:
+        # Get a copy of the current environment variables
+        env = os.environ.copy()
+
+        # Update the environment with custom variables from network_config
+        env.update({key.upper(): str(value) for key, value in deploy_config.items()})
+
+        proc.run(command.split(), env=env, cwd=project_dir or Path.cwd(), stdout_log_level=logging.INFO)
```
achidlow, line 116: if you make the suggested changes to `load_deploy_config` method, this method here could become trivial and maybe can be inlined instead, so the env manipulation can be done there

src/algokit/cli/deploy.py
aorumbayev: Done

src/algokit/cli/deploy.py
aorumbayev: Done 👍

src/algokit/cli/deploy.py
aorumbayev: I added an extra option to supply project directory, so if needed user can deploy from a different folder, that folder should contain algokit toml and env vars at the root

src/algokit/cli/deploy.py
aorumbayev: Done 👍

src/algokit/core/bootstrap.py
aorumbayev: Good catch, my vscode auto updated the import and i overlooked it, removed

src/algokit/core/bootstrap.py
aorumbayev: Done

src/algokit/core/constants.py
aorumbayev: There was some strange errors with mypy conflicting with ruff when using the union operator is the algorand networks, i've split them into 2 separate dics that are merged at the end, should be fixed now

src/algokit/core/deploy.py
aorumbayev: Removed

src/algokit/core/deploy.py
aorumbayev: done

src/algokit/core/deploy.py
aorumbayev: Added extra exception handling for key error and generic exception

src/algokit/core/deploy.py
aorumbayev: should be refined now

src/algokit/core/deploy.py
aorumbayev: It was required in the explore command as dapflow expects them in the url params and i haven't tweaked this constant when moving this to constants.py to be reused in other folders and its partially required when doing init on a typed client if lookup is done on creator and name. But otherwise by default it should not be required for lookup on ids so i'm setting ALGOD_SERVER as the only required default parameter for now, lmk what you think @achidlow

src/algokit/core/deploy.py
aorumbayev: I wanted to avoid implicit folder scanning as that can get a bit tricky since we always require the toml and env files to be placed on the same level, we also agreed on this convention with Alessandro Rob and Daniel so far but i agree that it needs some more flexibility to allow for deploying from custom paths. So:

1. .env files are now loaded as well IFF there is no env.network name prefixed files present 
2. if user needs deployment from other folders he can simply supply custom path via `--project-dir`

aorumbayev: Please resolve if you think this is a good alternative to walking up the dirs otherwise we can continue the thread @achidlow

src/algokit/core/deploy.py
aorumbayev: done

src/algokit/core/deploy.py
aorumbayev: done

src/algokit/core/deploy.py
aorumbayev: This would have been handy if i only needed to load the deploy config and pass it directly to the deploy commands, but if you check the lines 96-113 in deploy.py i also need extra work to be done there to identify if custom network being used for deployment isn't mainnet (if prod flag isn't passed, for extra security considerations it'll prompt for confirmation) additionally i am injecting the mnemonic and optional dispenser. 

Unless you meant something else here i think expecting the deploy config to be passed here instead of implicitly loading within method is still a bit cleaner

LMK what you think @achidlow

src/algokit/cli/deploy.py
aorumbayev: You are right, required isnt needed since we assume localnet by default - this should be fixed now and im using lower explicitly inside the function itself instead of the callback

src/algokit/cli/deploy.py
aorumbayev: replaced with getenv, as for "" vs None i believe it was just a typo

aorumbayev:
@achidlow phew most of the time spend was on approval tests to make sure everything is reverted back also is_localnet and is_mainnet are deadlocking when running in pytest so i had to mock them, given that these are unit tests i think i would have needed a mock there regardless but we should keep that in mind. 

Anyway i think your comments should be addressed now to recap:

- We agreed to revert the changes with consolidating stuff under constants file for now 
- Changes from your wip branch are merged here and further refined
- I got rid of the genesis call since we rely on is_localnet and is_mainnet now regardless so its not needed anymore
- Tests are updated for the latest state

In case you'd like any more tweaks before im back on Monday feel free to push directly to the branch before merging, also if i may ask - can you do a quick sanity check on the docs on the deploy command? I don't think our changes affect it in any way since its valid syntax for user regardless but you can push directly to the branch if you'll want to tweak/add anything on top

aorumbayev:
@achidlow @daniel-makerx feel free to check again, also not sure if we want to use the `lambda: "CI" not in os.environ` , github actions passes this value as true by default, so user needs to be aware of any unexpected behaviour in case these overlap. For now monkeypatching the CI value to make sure some tests on bootstrap and deploy work fine when running from ci

src/algokit/cli/deploy.py
```diff
@@ -0,0 +1,104 @@
...
+)
+def deploy_command(
+    *,
+    environment_name: str | None,
+    command: list[str] | None,
+    interactive: bool,
+    path: Path,
+) -> None:
+    """Deploy smart contracts from AlgoKit compliant repository."""
+    if environment_name:
```
achidlow, line 70: if we lower case and the user has any uppercase characters and is using a case-sensitive file system, this means they won't be able to use that file

aorumbayev: @achidlow refined docs to mention that point and removed case insensitivity

docs/features/deploy.md
```diff
@@ -0,0 +1,130 @@
...
+[deploy]
+command = "your command"
+environment_secrets = [
+  "YOUR_COMMON_SECRET",
+]
+
+[deploy.localnet]
+command = "your localnet-specific command"  # Optional. If not specified, the generic command from [deploy] will be used.
+environment_secrets = [
+  "YOUR_LOCALNET_SECRET",
```
daniel-makerx, line 64: I think it might be worth simplifying this example? While this will work, it probably isn't a typical configuration. Maybe have something similar to the template, where we have some common secrets and then override those secrets in localnet as they aren't required.

docs/features/deploy.md
```diff
@@ -0,0 +1,130 @@
...
+$ algokit deploy testnet --ci
+```
+
+## Example of a Full Deployment
+
+```sh
+$ algokit deploy testnet --custom-deploy-command="your-custom-command"
+```
+
+This example shows how to deploy smart contracts to the testnet using a custom deploy command and skipping the warning prompt. This also assumes that .algokit.toml file is present in the current working directory, and .env.testnet file is present in the current working directory and contains the required environment variables for deploying to TestNet environment.
```
daniel-makerx, line 126: Is the bit about a warning prompt still relevant?

aorumbayev: good catch, removed the part mentioning prompt

docs/features/deploy.md
aorumbayev: Refined to include the base sample from the beaker template (deploy branch )

docs/features/deploy.md
```diff
@@ -0,0 +1,127 @@
...
+The logic for loading environment variables is as follows:
+
+- If a `.env` file exists, the environment variables contained in it are loaded first.
+- If a `.env.[network_name]` file exists, the environment variables in it are loaded, overriding any previously loaded values from the `.env` file for the same variables.
+
+## AlgoKit Configuration File
+
+AlgoKit uses a configuration file called `.algokit.toml` in the root of your project. The configuration file can be created using the `algokit init` command. This file will define the deployment commands for the various network environments that you want to target.
+
+Here's an example of what the `.algokit.toml` file might look like. The example below will always prompt for the deployer mnemonic secret (unless specified in env variables) for any network, except for the localnet, where it will not prompt for any secrets.
```
daniel-makerx, line 49:
```suggestion
Here's an example of what the `.algokit.toml` file might look like. When deploying it will prompt for the `DEPLOYER_MNEMONIC` secret unless it is already defined as an environment variable or is deploying to localnet.
```

PR #306, opened 2023-07-26
feat: advanced algokit generate command

aorumbayev:
## Proposed changes

- Adding logic to dynamically load new commands into click based on algokit.toml file. This is achieved by overloading Group class on click.
- New commands are referred to custom generators that can optionally exist on projects created by algokit tempaltes

## Further notes

Do not merge until 1.3.0 is released first


src/algokit/cli/generate.py
```diff
@@ -3,12 +3,36 @@
...
+
+    def list_commands(self, ctx: click.Context) -> list[str]:
+        predefined_command_names = super().list_commands(ctx)
+        dynamic_commands = load_custom_generate_commands(Path.cwd())
+        dynamic_command_names = [command.name for command in dynamic_commands if command and command.name]
+
+        return sorted(predefined_command_names + dynamic_command_names)
+
+
+@click.group("generate", cls=GeneratorGroup)
```
robdmoore, line 35: Cool!

src/algokit/core/generate.py
```diff
@@ -0,0 +1,121 @@
...
+def load_custom_generate_commands(project_dir: Path) -> list[click.Command]:
+    generators = _load_generators(project_dir)
+    commands = []
+
+    for generator in generators:
+        # Set the generator name and description as a new command
+        command = click.command(name=generator.name, help=generator.description)(_run_generator)
+
+        # Add the options to the command to allow passing answers and custom path (optional)
+        command = click.option(
```
robdmoore, line 101: Is it possible to output the answers for that template for the help documentation or it is too hard?

tests/generate/test_generate_generators.test_generate_generators_valid_generator_run.approved.txt
```diff
@@ -0,0 +1 @@
+DEBUG: Running generator in <MagicMock name='Worker().__enter__().src_path' id='4383216400'>
```
robdmoore, line 1: Is this right?

tests/generate/test_generate_generators.test_generate_generators_valid_generator_invalid_path.approved.txt
```diff
@@ -0,0 +1,9 @@
+Usage: algokit generate [OPTIONS] COMMAND [ARGS]...
```
robdmoore, line 1: If the path is invalid should there be some difference we expect?

tests/generate/test_generate_generators.test_generate_generators_valid_generator_no_description.approved.txt
```diff
@@ -0,0 +1,9 @@
+Usage: algokit generate [OPTIONS] COMMAND [ARGS]...
```
robdmoore, line 1: This looks the same as the others, do we expect a difference?

tests/generate/test_generate_generators.test_generate_generators_no_toml.approved.txt
```diff
@@ -0,0 +1,9 @@
+Usage: algokit generate [OPTIONS] COMMAND [ARGS]...
```
robdmoore, line 1: Should there be some difference when there is no toml?

src/algokit/core/generate.py
aorumbayev: @robdmoore it's already the case, the `description` of the generation (if present in the toml) is what being mapped to the help of the command. See line 98

tests/generate/test_generate_generators.test_generate_generators_valid_generator_no_description.approved.txt
aorumbayev: @robdmoore this was also an artifact from the test file before it was renamed, see all tests under test_generate_custom_generate_commands.* prefix

tests/generate/test_generate_generators.test_generate_generators_no_toml.approved.txt
aorumbayev: @robdmoore  see all tests under test_generate_custom_generate_commands.* prefix

src/algokit/cli/generate.py
```diff
@@ -3,12 +3,36 @@
...
+class GeneratorGroup(click.Group):
+    def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None:
+        return_value = super().get_command(ctx, cmd_name)
+
+        if return_value is not None:
+            return return_value
+
+        commands = load_custom_generate_commands(Path.cwd())
+
+        for command in commands:
```
achidlow, line 21: what about making `load_custom_generate_commands` return mapping str -> Command, then you can just `commands.get(cmd_name)`?

src/algokit/cli/generate.py
```diff
@@ -3,12 +3,36 @@
...
+            if command.name == cmd_name:
+                return command
+
+        return None
+
+    def list_commands(self, ctx: click.Context) -> list[str]:
+        predefined_command_names = super().list_commands(ctx)
+        dynamic_commands = load_custom_generate_commands(Path.cwd())
+        dynamic_command_names = [command.name for command in dynamic_commands if command and command.name]
```
achidlow, line 30: similarly, this could then just be `dynamic_command_names = list(dynamic_commands)`

src/algokit/core/generate.py
```diff
@@ -0,0 +1,123 @@
...
+
+    generators_table = config.get("generate", {})
+
+    if not isinstance(generators_table, dict):
+        raise click.ClickException(f"Bad data for [generators] key in '{ALGOKIT_CONFIG}'")
+
+    for name, generators_config in generators_table.items():
+        match generators_config:
+            case dict():
+                if not generators_config.get("path"):
```
achidlow, line 73: you can split this one into a separate case, or rather split the successful case:
`case {"path": str(path), **remaining}`

src/algokit/core/generate.py
```diff
@@ -0,0 +1,123 @@
...
+                    logger.warning(f"Missing path for generator '{name}' in '{ALGOKIT_CONFIG}', skipping")
+                    continue
+                if not Path.exists(Path(generators_config.get("path", ""))):
```
achidlow, line 76: why `Path.exists(Path(foo))` instead of just `Path(foo).exists()`?

achidlow: Also, in case you don't take the suggestion of splitting the successful case above, maybe just assign `generators_config.get("path")` to a variable above and below within this case?

src/algokit/core/generate.py
```diff
@@ -0,0 +1,123 @@
...
+                    )
+                    continue
+
+                generator = Generator(
+                    name=_format_generator_name(name),
+                    description=generators_config.get("description"),
+                    path=generators_config.get("path"),
+                )
+                generators.append(generator)
+            case str():
```
achidlow, line 88: should this just be `case _:` ?

src/algokit/core/generate.py
```diff
@@ -0,0 +1,123 @@
...
+                logger.debug(f'Invalid generator configuration key "{name}" of value "{generators_config}", skipping')
+
+    return generators
+
+
+def load_custom_generate_commands(project_dir: Path) -> list[click.Command]:
```
achidlow, line 94: I think this function should be under cli/generate.py

I know the boundaries are a bit loose, but generally anything the concerns the "UI" should go in the cli part, and stuff not related so much to click should be in core. We have probably been a bit lax already and violated this concept in a few places, but I would like to avoid doing so any further.

src/algokit/core/generate.py
```diff
@@ -0,0 +1,123 @@
...
+    generators = _load_generators(project_dir)
+    commands = []
+
+    for generator in generators:
+        # Set the generator name and description as a new command
+        command = click.command(name=generator.name, help=generator.description)(_run_generator)
```
achidlow, line 100: would it be simpler to set this function up along with it's associated decorators using actual function syntax here instead of composing decorators manually?

e.g.

```py3
@click.command(...)
@click.option(...)
def command(...):
    return _run_generator()
```

tests/generate/test_generate_custom_generate_commands.py
```diff
@@ -0,0 +1,125 @@
...
+    verify(result.output)
+
+
+def test_generate_custom_generate_commands_valid_generator(
+    tmp_path_factory: TempPathFactory,
+) -> None:
+    cwd = tmp_path_factory.mktemp("cwd")
+    (cwd / "smart_contract").mkdir()
+    # Required for windows compatibility
+    smart_contract_path = str((cwd / "smart_contract").absolute()).replace("\\", r"\\")
```
achidlow, line 47: the above 3 lines are copy pasted a few times, maybe have a fixture and/or helper function? we don't need to overthink / over engineer the tests, but if a single helper function aids readability and understanding then we should do that.

src/algokit/cli/generate.py
aorumbayev: @achidlow good point, refined to return a dict mapping

src/algokit/cli/generate.py
aorumbayev: @achidlow done 👍

src/algokit/core/generate.py
aorumbayev: @achidlow ended up going the later route by splitting into a successfull case.

src/algokit/core/generate.py
aorumbayev: @achidlow done 👍

src/algokit/core/generate.py
aorumbayev: @achidlow done 👍

src/algokit/core/generate.py
aorumbayev: @achidlow thanks for clarifying - moved it to the cli side, its now a private method. Private methods in core are not exported and public

tests/generate/test_generate_custom_generate_commands.py
aorumbayev: @achidlow done - added a simple fixture to encapsulate these 3 lines and return cwd, custom_folder_path as a tuple👍

src/algokit/cli/generate.py
```diff
@@ -3,12 +3,72 @@
...
+            help="Answers key/value pairs to pass to the template.",
+            nargs=2,
+            default=[],
+            metavar="<key> <value>",
+        )
+        @click.option(
+            "path",
+            "--path",
+            "-p",
+            help=f"Path to {generator.name} generator. (Default: {generator.description})",
```
daniel-makerx, line 41: should the default refer to {generator.path} ?

src/algokit/core/generate.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Run the custom generator with the given answers and path.
+    :param answers: Answers to pass to the generator.
+    :param path: Path to the generator.
+    """
+
+    # copier is lazy imported for two reasons
+    # 1. it is slow to import on first execution after installing
+    # 2. the import fails if git is not installed (which we check above)
+    from copier.main import Worker  # type: ignore[import]
```
daniel-makerx, line 39: This (or perhaps in the cli layer) should check for the git command, and raise a helpful error if it isn't present. See init command here https://github.com/algorandfoundation/algokit-cli/blob/main/src/algokit/cli/init.py#L188

src/algokit/cli/generate.py
aorumbayev: @daniel-makerx just pushed a fix, thanks for catching this typo

src/algokit/core/generate.py
aorumbayev: @daniel-makerx added a check on cli side before running the run generator in core. And corresponding test let me know if there is anything else

daniel-makerx (approved):

robdmoore (approved):

docs/features/generate.md
```diff
@@ -70,22 +59,109 @@ response = app_client.hello(name="World")
 
 ```typescript
 // A similar working example can be seen in the beaker_production template, when using TypeScript deployment
-import { HelloWorldAppClient } from './artifacts/HelloWorldApp/client'
+import { HelloWorldAppClient } from "./artifacts/HelloWorldApp/client";
```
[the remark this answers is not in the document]
aorumbayev: @robdmoore prettier auto changed this to default settings, will address it in separate pr (potentially moving prettierrc to the root of the repo). Only restoring this code snippet to initial state within this pr scope

PR #308, opened 2023-08-08
fix: localnet displays a warning when image is out of date

inaie-makerx:
https://dev.azure.com/MakerXStudio/AlgoKit%20(Algorand%20Foundation)/_workitems/edit/431

### DIFF
--- a/src/algokit/core/sandbox.py
+++ b/src/algokit/core/sandbox.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from typing import Any, cast
 
 import httpx
-import yaml
 
 from algokit.core.conf import get_app_config_dir
 from algokit.core.proc import RunResult, run, run_interactive
--- a/src/algokit/core/sandbox.py
+++ b/src/algokit/core/sandbox.py
@@ -120,18 +120,20 @@ class ComposeSandbox:
         assert isinstance(data, list)
         return cast(list[dict[str, Any]], data)
 
-    def _get_local_image_version(self, image_name: str) -> str:
+    def _get_local_image_version(self, image_name: str) -> str | None:
         """
         Get the local version of a Docker image
         """
-        arg = "{{.RepoDigests}}"
-        local_version = run(
-            ["docker", "image", "inspect", image_name, "--format", arg],
-            cwd=self.directory,
-            bad_return_code_error_message="Failed to get image inspect",
-        )
-        # Remove brackets and split on '@' symbol and get the SHA hash
-        return local_version.output[1:-1].split("@")[1]
+        try:
+            arg = "{{.Id}}"
+            local_version = run(
+                ["docker", "image", "inspect", image_name, "--format", arg],
+                cwd=self.directory,
+                bad_return_code_error_message="Failed to get image inspect",
+            )
+            return local_version.output
+        except Exception:
+            return None
 
     def _get_latest_image_version(self, image_name: str) -> str | None:
         """
@@ -151,19 +153,19 @@ class ComposeSandbox:
     def is_image_up_to_date(self, image_name: str) -> bool:
         local_version = self._get_local_image_version(image_name)
         latest_version = self._get_latest_image_version(image_name)
-        return local_version is None or local_version == latest_version
+        return local_version is None or latest_version is None or local_version == latest_version
 
     def check_docker_compose_for_new_image_versions(self) -> None:
         is_indexer_up_to_date = self.is_image_up_to_date(INDEXER_IMAGE)
         if is_indexer_up_to_date is False:
             logger.warning(
-                "indexer has a new version available, run algokit localnet reset --update to get the latest version"
+                "indexer has a new version available, run `algokit localnet reset --update` to get the latest version"
             )
 
         is_algorand_up_to_date = self.is_image_up_to_date(ALGORAND_IMAGE)
         if is_algorand_up_to_date is False:
             logger.warning(
-                "algod has a new version available, run algokit localnet reset --update to get the latest version"
+                "algod has a new version available, run `algokit localnet reset --update` to get the latest version"
             )
 
 
--- a/src/algokit/core/sandbox.py
+++ b/src/algokit/core/sandbox.py
@@ -134,7 +134,7 @@ class ComposeSandbox:
         # Remove brackets and split on '@' symbol and get the SHA hash
         return local_version.output[1:-1].split("@")[1]
 
-    def _get_latest_image_version(self, image_name: str) -> str:
+    def _get_latest_image_version(self, image_name: str) -> str | None:
         """
         Get the latest version of a Docker image from Docker Hub
         """
@@ -147,22 +147,25 @@ class ComposeSandbox:
             return str(data.json()["digest"])
         except Exception as err:
             logger.debug(f"Error checking indexer status: {err}", exc_info=True)
-            return ""
+            return None
+
+    def is_image_up_to_date(self, image_name: str) -> bool:
+        local_version = self._get_local_image_version(image_name)
+        latest_version = self._get_latest_image_version(image_name)
+        return local_version is None or local_version == latest_version
 
     def check_docker_compose_for_new_image_versions(self) -> None:
-        """
-        Check Docker Compose file for new image versions
-        """
-        docker_compose = yaml.safe_load(get_docker_compose_yml())
-        services = docker_compose["services"]
-        for service_name, service_data in services.items():
-            image_name = service_data["image"]
-            if "postgres" in image_name:
-                continue
-            local_version = self._get_local_image_version(image_name)
-            latest_version = self._get_latest_image_version(image_name)
-            if local_version != latest_version:
-                logger.warning(f"{service_name} has a new version available")
+        is_indexer_up_to_date = self.is_image_up_to_date(INDEXER_IMAGE)
+        if is_indexer_up_to_date is False:
+            logger.warning(
+                "indexer has a new version available, run algokit localnet reset --update to get the latest version"
+            )
+
+        is_algorand_up_to_date = self.is_image_up_to_date(ALGORAND_IMAGE)
+        if is_algorand_up_to_date is False:
+            logger.warning(
+                "algod has a new version available, run algokit localnet reset --update to get the latest version"
+            )
 
 
 DEFAULT_ALGOD_SERVER = "http://localhost"
@@ -172,6 +175,8 @@ DEFAULT_INDEXER_PORT = 8980
 DEFAULT_WAIT_FOR_ALGOD = 60
 DEFAULT_HEALTH_TIMEOUT = 1
 ALGOD_HEALTH_URL = f"{DEFAULT_ALGOD_SERVER}:{DEFAULT_ALGOD_PORT}/v2/status"
+INDEXER_IMAGE = "makerxau/algorand-indexer-dev:latest"
+ALGORAND_IMAGE = "algorand/algod:latest"
 
 
 def _wait_for_algod() -> bool:
@@ -215,7 +220,7 @@ name: "{name}_sandbox"
 services:
   algod:
     container_name: {name}_algod
-    image: algorand/algod:latest
+    image: {ALGORAND_IMAGE}
     ports:
       - {algod_port}:8080
       - {kmd_port}:7833
@@ -232,7 +237,7 @@ services:
 
   indexer:
     container_name: {name}_indexer
-    image: makerxau/algorand-indexer-dev:latest
+    image: {INDEXER_IMAGE}
     ports:
       - {indexer_port}:8980
     restart: unless-stopped
--- a/src/algokit/core/sandbox.py
+++ b/src/algokit/core/sandbox.py
@@ -1,11 +1,14 @@
 import enum
 import json
 import logging
+import subprocess
 import time
 from pathlib import Path
 from typing import Any, cast
 
 import httpx
+import requests
+import yaml
 
 from algokit.core.conf import get_app_config_dir
 from algokit.core.proc import RunResult, run, run_interactive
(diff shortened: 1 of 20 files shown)


src/algokit/core/sandbox.py
```diff
@@ -120,6 +121,49 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
...
+            return str(data.json()["digest"])
+        except Exception as err:
+            logger.debug(f"Error checking indexer status: {err}", exc_info=True)
+            return ""
+
+    def check_docker_compose_for_new_image_versions(self) -> None:
+        """
+        Check Docker Compose file for new image versions
+        """
+        docker_compose = yaml.safe_load(get_docker_compose_yml())
```
daniel-makerx, line 156: Rather than loading the docker-compose.yaml contents, could instead make the algod and indexer image tags constants and reference them here and in the `get_docker_compose_yml`

src/algokit/core/sandbox.py
```diff
@@ -120,6 +121,49 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
...
+        services = docker_compose["services"]
+        for service_name, service_data in services.items():
+            image_name = service_data["image"]
+            if "postgres" in image_name:
+                continue
+            local_version = self._get_local_image_version(image_name)
+            latest_version = self._get_latest_image_version(image_name)
+            if local_version != latest_version:
+                logger.warning(f"{service_name} has a new version available")
```
daniel-makerx, line 165: Include in this message what the user has to do to get the latest version (run with --update)

src/algokit/core/sandbox.py
```diff
@@ -120,6 +121,49 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
...
+        args = image_name.split(":")
+        name = args[0]
+        tag = args[1] if len(args) > 1 else "latest"
+        url = f"https://registry.hub.docker.com/v2/repositories/{name}/tags/{tag}"
+        try:
+            data = httpx.get(url=url)
+            return str(data.json()["digest"])
+        except Exception as err:
+            logger.debug(f"Error checking indexer status: {err}", exc_info=True)
+            return ""
```
daniel-makerx, line 150: I think this should return None here if we can't fetch the latest, and then skip the warning if None is returned so that users don't get warnings if we can't determine the latest version

tests/localnet/conftest.py
```diff
@@ -13,3 +18,57 @@ def algod_health_fast_timings(mocker: MockerFixture) -> None:  # noqa: ignore[PT
...
+    httpx_mock.add_response(
+        url="https://registry.hub.docker.com/v2/repositories/algorand/algod/tags/latest",
+        json={
+            "digest": "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+        },
+    )
+
+
+@pytest.fixture()
+def localnet_out_date(proc_mock: ProcMock, httpx_mock: HTTPXMock) -> None:
```
daniel-makerx, line 51: `localnet_out_of_date` ?

tests/localnet/conftest.py
```diff
@@ -13,3 +18,57 @@ def algod_health_fast_timings(mocker: MockerFixture) -> None:  # noqa: ignore[PT
 @pytest.fixture()
 def health_success(httpx_mock: HTTPXMock) -> None:  # noqa: ignore[PT004]
     httpx_mock.add_response(url=ALGOD_HEALTH_URL)
+
+
+@pytest.fixture()
+def localnet_up_to_date(proc_mock: ProcMock, httpx_mock: HTTPXMock) -> None:
+    proc_mock.set_output(
+        ["docker", "image", "inspect", "algorand/algod:latest", "--format", "{{.RepoDigests}}"],
```
daniel-makerx, line 26: If we make the image tags constants as mentioned earlier, then they can also be referenced here in the tests instead of defining them in multiple places

tests/localnet/conftest.py
```diff
@@ -13,3 +15,57 @@ def algod_health_fast_timings(mocker: MockerFixture) -> None:  # noqa: ignore[PT
 @pytest.fixture()
 def health_success(httpx_mock: HTTPXMock) -> None:  # noqa: ignore[PT004]
     httpx_mock.add_response(url=ALGOD_HEALTH_URL)
+
+
+@pytest.fixture()
+def _localnet_up_to_date(proc_mock: ProcMock, httpx_mock: HTTPXMock) -> None:
```
aorumbayev, line 21: Minor detail, i think
`_localnet_up_to_date_` -> `localnet_up_to_date` might be better? 

Its used in at least 2 separate test sets so i assume it makes more sense to remove the `_`. If you look at the only similar example of 'private' fixture in `test_doctor.py` that fixture is only used within a scope of a single file hence makes sense to mark it private there.

tests/localnet/conftest.py
```diff
@@ -13,3 +15,57 @@ def algod_health_fast_timings(mocker: MockerFixture) -> None:  # noqa: ignore[PT
...
+    httpx_mock.add_response(
+        url="https://registry.hub.docker.com/v2/repositories/algorand/algod/tags/latest",
+        json={
+            "digest": "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+        },
+    )
+
+
+@pytest.fixture()
+def _localnet_out_of_date(proc_mock: ProcMock, httpx_mock: HTTPXMock) -> None:
```
aorumbayev, line 48: In contrast with `localnet_up_to_date`, `_localnet_out_of_date` is only used by a single test case in `test_localnet_start.py`. Is there any need to have it defined in conftest in this case?

src/algokit/core/sandbox.py
```diff
@@ -120,6 +120,52 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
...
+        is_indexer_up_to_date = self.is_image_up_to_date(INDEXER_IMAGE)
+        if is_indexer_up_to_date is False:
+            logger.warning(
+                "indexer has a new version available, run algokit localnet reset --update to get the latest version"
+            )
+
+        is_algorand_up_to_date = self.is_image_up_to_date(ALGORAND_IMAGE)
+        if is_algorand_up_to_date is False:
+            logger.warning(
+                "algod has a new version available, run algokit localnet reset --update to get the latest version"
```
aorumbayev, line 166:
```suggestion
                "algod has a new version available, run `algokit localnet reset --update` to get the latest version"
```

src/algokit/core/sandbox.py
```diff
@@ -120,6 +120,52 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
...
+    def is_image_up_to_date(self, image_name: str) -> bool:
+        local_version = self._get_local_image_version(image_name)
+        latest_version = self._get_latest_image_version(image_name)
+        return local_version is None or local_version == latest_version
+
+    def check_docker_compose_for_new_image_versions(self) -> None:
+        is_indexer_up_to_date = self.is_image_up_to_date(INDEXER_IMAGE)
+        if is_indexer_up_to_date is False:
+            logger.warning(
+                "indexer has a new version available, run algokit localnet reset --update to get the latest version"
```
aorumbayev, line 160:
```suggestion
                "indexer has a new version available, run `algokit localnet reset --update` to get the latest version"
```

src/algokit/core/sandbox.py
```diff
@@ -120,6 +120,52 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
...
+        Get the local version of a Docker image
+        """
+        arg = "{{.RepoDigests}}"
+        local_version = run(
+            ["docker", "image", "inspect", image_name, "--format", arg],
+            cwd=self.directory,
+            bad_return_code_error_message="Failed to get image inspect",
+        )
+        # Remove brackets and split on '@' symbol and get the SHA hash
+        return local_version.output[1:-1].split("@")[1]
```
aorumbayev, line 134: output has a newline character at the end which isn't accounted for, can result in a bug where it forgets to strip the rightmost `]` symbol. See if there are better ways to get just the sha part, maybe --format flag has some further ways to parse the output on docker side instead? Otherwise make sure to run strip() on the output before further operations

aorumbayev: Please also make sure to also reflect the real outputs in the mocks, current tests aren't catching scenarios like that because mocked return values are already trimmed. Potentially adding extra asserts where applicable to ensure the sha strings were correctly parsed and etc (there are no such test checks at the moment)

tests/localnet/test_localnet_start.py
```diff
@@ -168,11 +169,19 @@ def test_localnet_start_with_unparseable_docker_compose_version(app_dir_mock: Ap
...
     proc_mock.set_output("docker compose version --format json", [json.dumps({"version": "v2.10.0-gitpod.0"})])
 
     result = invoke("localnet start")
 
     assert result.exit_code == 0
     verify(result.output.replace(str(app_dir_mock.app_config_dir), "{app_config}").replace("\\", "/"))
+
+
+@pytest.mark.usefixtures("proc_mock", "health_success", "_localnet_out_of_date")
+def test_localnet_start_out_date(app_dir_mock: AppDirs) -> None:
```
aorumbayev, line 183: Might be useful to add few more tests testing edge cases when you have local docker inspect invocation failing on some exception (should be handled without raising any unhandled exceptions). Similarly testing a scenario where remote local inspect throws an error, as expected though it should print a debug warning but user of the cli shouldn't see any errors/logs and instead its just continues with the execution of the command.

aorumbayev (changes requested):

tests/localnet/conftest.py
inaie-makerx: I added the underscore because of this error 
Fixture `localnet_up_to_date` does not return anything, add leading underscoreRuff(PT004)

src/algokit/core/sandbox.py
inaie-makerx: I have changed to use the .Id it seems to return the sha formatted on the way we need already

tests/localnet/conftest.py
aorumbayev: I see, looks like ruff is enforcing pytest conventions, ignore this one then all good

src/algokit/core/sandbox.py
aorumbayev: @inaie-makerx I'm sorry you can't really use ID in this case case RepoDigest vs ID are derived in different manners and we rely on repo digest on remote image side. All i meant was that when you run algokit cli (aside from mocks in the tests) this inspect response returns a newline character (happens in current state with ID as well). 

Can you please:
1. Make sure the output is stripper/trimmed from any special characters on left/right side before you return it from `_get_local_image_version` ?
2. Change the ID back to RepoDigest? (see [ref](https://stackoverflow.com/questions/56364643/whats-the-difference-between-a-docker-images-image-id-and-its-digest))

Here is what you can do as an option, docker relies on go under the hood so you can rely on go templating capabilities to do something like this:

```python
    def _get_local_image_version(self, image_name: str) -> str | None:
        """
        Get the local version of a Docker image
        """
        try:
            arg = '{{index (split (index .RepoDigests 0) "@") 1}}'
            local_version = run(
                ["docker", "image", "inspect", image_name, "--format", arg],
                cwd=self.directory,
                bad_return_code_error_message="Failed to get image inspect",
            )
            # Remove brackets and split on '@' symbol and get the SHA hash
            return local_version.output.strip()
        except Exception:
            return None
```

aorumbayev (changes requested):
@inaie-makerx unresolved one of the comments as the bug still persists, please address and I'll approve if all good.

tests/localnet/test_localnet_start.py
```diff
@@ -12,7 +18,42 @@
 from tests.utils.proc_mock import ProcMock
 
 
-@pytest.mark.usefixtures("proc_mock", "health_success")
+@pytest.fixture()
+def _localnet_out_of_date(proc_mock: ProcMock, httpx_mock: HTTPXMock) -> None:
+    arg = '{{index (split (index .RepoDigests 0) "@") 1}}'
+    proc_mock.set_output(
+        ["docker", "image", "inspect", ALGORAND_IMAGE, "--format", arg],
+        ["[algorand/algod@sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb\n]"],
```
aorumbayev, line 26: Should be just `sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb\n` without square brackets due to `{{index (split (index .RepoDigests 0) "@") 1}}`

tests/localnet/test_localnet_start.py
```diff
@@ -12,7 +18,42 @@
...
+    )
+
+    proc_mock.set_output(
+        ["docker", "image", "inspect", INDEXER_IMAGE, "--format", arg],
+        ["[makerxau/algorand-indexer-dev@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n]"],
```
aorumbayev, line 31: Same here,  should be just `sha256:aa....\n`

aorumbayev (approved):
@inaie-makerx approving but please address the last comment with regards to values you are mocking for inspect command. Before merging, please make sure that conventional commit prefix is set to `fix` instead of a `feat` as this is more of a minor enhancement into existing localnet features so a patch bump is more appropriate than a minor bump. 

@daniel-makerx any chance you can take one extra look as well so we get second approval for this PR?

robdmoore (approved):

robdmoore:
Nice!

tests/localnet/test_localnet_start.py
aorumbayev: These are still in place? Am i missing something? I think you pushed a commit to remove them from the test but unsure if these are still correct, the output should not contain [{title}/{image title}@sha256:{hash}] = we parse it out via go templating in format param.

tests/localnet/test_localnet_start.py
aorumbayev: These are still in place? Am i missing something? I think you pushed a commit to remove them from the test but unsure if these are still correct, the output should not contain [{title}/{image title}@sha256:{hash}] = we parse it out via go templating in format param.

PR #313, opened 2023-09-25
feat: algokit task - wallet aliasing

aorumbayev:
## Proposed Changes

  - Introduces wallet aliasing task 
  - Supports add, list, remove, reset, get operations 
  - Capping max number of aliases to have in keyring at a time to 50 (as discussed with @Loedn)
 
## TODO

- [x] snapshot tests

## Further notes

Refactoring deploy command to support aliases are to be addressed in a separate PR


aorumbayev:
@neilcampbell @robdmoore @Loedn, update so apparently max password len on windows is 1280 chars which is roughly ~60 aliases assuming worst case when max alias len used is 20 chars. I think 60 aliases is pretty reasonable amount - are we ok with throwing a warning that max number of aliases to persist is exhausted or we want unlimited storage for aliases? Storing it in keyring might get a bit tricky in case we want to remove limits

src/algokit/cli/tasks/wallet.py
```diff
@@ -0,0 +1,165 @@
...
+        }
+        for alias_data in aliases
+    ]
+
+    click.echo(json.dumps(output, indent=2))
+
+
+@wallet.command("remove")
+@click.argument("alias", type=click.STRING)
+def remove(alias: str) -> None:
```
neilcampbell, line 125: This is also a destructive action, so we probably want to prompt for confirmation and add a --force flag too.

docs/features/tasks/wallet.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Add
+
+This command adds an address or account to be stored against a named alias. If the `--mnemonic` flag is used, it will prompt the user for a mnemonic phrase interactively using masked input. If the `--force` flag is used, it will allow overwriting an existing alias.
+
+```bash
+$ algokit wallet add [OPTIONS] ALIAS_NAME
+```
+
+> Please note, the command is not designed to be used in CI scope, there is not options to skip interactive masked input of mnemonic if you want to alias an Account (both private and public key) entity.
```
robdmoore, line 36:
```suggestion
> Please note, the command is not designed to be used in CI scope, there is no option to skip interactive masked input of mnemonic if you want to alias an 'Account' (both private and public key) entity.
```

src/algokit/core/tasks/wallet.py
```diff
@@ -0,0 +1,165 @@
...
+from dataclasses import dataclass
+
+import keyring
+
+logger = logging.getLogger(__name__)
+
+WALLET_ALIAS_KEYRING_NAMESPACE = "algokit_alias"
+WALLET_ALIASES_KEYRING_NAMESPACE = "algokit_aliases"
+WALLET_ALIASES_KEYRING_USERNAME = "aliases"
+WALLET_ALIASING_MAX_LIMIT = 50
```
robdmoore, line 12: Minor: Worth including a description of why this is this number? I assume it's a keyring limitation?

src/algokit/cli/tasks/wallet.py
```diff
@@ -0,0 +1,165 @@
...
+        derived_address = account.address_from_private_key(private_key)  # type: ignore[no-untyped-call]
+        if derived_address != address:
+            click.echo(
+                "Warning: Address from the mnemonic doesn't match the provided address. "
+                "It won't work unless the account has been rekeyed."
+            )
+
+    if get_alias(alias_name) and not force:
+        response = click.prompt(
+            f"Alias '{alias_name}' already exists. Overwrite? (y/n)",
```
neilcampbell, line 73:
```suggestion
            f"Alias '{alias_name}' already exists. Overwrite?",
```

Not needed.

src/algokit/core/tasks/wallet.py
```diff
@@ -0,0 +1,165 @@
...
+
+    try:
+        response = keyring.get_password(service_name=WALLET_ALIAS_KEYRING_NAMESPACE, username=alias_name)
+
+        if not response:
+            return None
+
+        return WalletAliasKeyringData(**json.loads(response))
+    except Exception as ex:
+        logger.debug(f"Failed to get alias from keyring. `{alias_name}` does not exist", exc_info=ex)
```
neilcampbell, line 129:
```suggestion
        logger.debug(f"`{alias_name}` does not exist", exc_info=ex)
```
Based on snapshot test, this appears to be the way keyring signals that a value doesn't exist. Remove failure wording, as we expect these messages reqgularly.

neilcampbell (approved):

PR #314, opened 2023-09-26
feat: algokit task - vanity address

negar-abbasi:
## Proposed Changes

  - Introduces vanity-address task 
 
## TODO

- [x] more test cases
- [x] add more explanation on generating mnemonic on stdout

## Further notes 

Has to merged AFTER https://github.com/algorandfoundation/algokit-cli/pull/313


src/algokit/cli/tasks/vanity_address.py
```diff
@@ -0,0 +1,147 @@
...
+    lock: LockBase,
+    stop_event: EventClass,
+    shared_dict: DictProxy,
+) -> None:
+    last_log_time = start_time
+    while not stop_event.is_set():
+        private_key, address = algosdk.account.generate_account()  # type: ignore[no-untyped-call]
+        generated_mnemonic = mnemonic.from_private_key(private_key)  # type: ignore[no-untyped-call]
+        with lock:
+            shared_dict["count"] += 1
```
robdmoore, line 119: Minor (not needed to merge PR): I just ran a similar script locally and I have a suspicion that locking on the count is a bottleneck that slows things down. As a suggestion, if we instead have an array of values so each process has it's own count then it wouldn't need to lock. Alternatively, if each process detects the stop event and pushes a count it's maintaining in it's own thread to a queue or whatever that would be another way to avoid locking and speed things up.

aorumbayev: @robdmoore the bottleneck was in the fact that the script wasn't using multithreading at all, pushed a set of refinements please check again. Also fyi this is lower priority, we would like to merge the aliasing first -> https://github.com/algorandfoundation/algokit-cli/pull/313

src/algokit/core/tasks/vanity_address.py
```diff
@@ -0,0 +1,108 @@
...
+        private_key, address = algosdk.account.generate_account()  # type: ignore[no-untyped-call]
+        generated_mnemonic = mnemonic.from_private_key(private_key)  # type: ignore[no-untyped-call]
+
+        match_conditions = {
+            MatchType.START: address.startswith(keyword),
+            MatchType.ANYWHERE: keyword in address,
+            MatchType.END: address.endswith(keyword),
+        }
+
+        with lock:
```
robdmoore, line 66: Note: I had a minor performance related comment on this line in the other PR

robdmoore (approved):

src/algokit/core/tasks/vanity_address.py
aorumbayev: @robdmoore refactored further to get rid of locking, instead there is a dedicated counter and the thread that runs the logger sums up those counters periodically

PR #324, opened 2023-10-09
feat: transaction signing task; polishing tasks

aorumbayev:
## Proposed changes

- This PR refactors the code by moving some functions from transfer.py to utils.py for better code organization and reusability. It also adds a new feature to sign Algorand transactions in sign_transaction.py.
- Adding `.` at the end of all help messages in tasks for consistency. Updating affected tests, regenerating docs

## TODO

- [x] Snapshot tests

## Open questions

- Acceptance criteria explicitly mentions that this operation allows working with single transaction per file/base64 encoded record passed. @robdmoore just making sure we are on the same page, as you can see in the implementation if user provides a file with more than 1 txn in the file click will raise an exception saying only 1 txn per file is supported for signing

### DIFF
--- a/src/algokit/cli/tasks/sign_transaction.py
+++ b/src/algokit/cli/tasks/sign_transaction.py
@@ -1,63 +1,121 @@
+import base64
 import json
+import logging
 from pathlib import Path
 from typing import cast
 
 import click
 from algosdk import encoding
-from algosdk.transaction import Transaction, retrieve_from_file, write_to_file
+from algosdk.transaction import SignedTransaction, Transaction, retrieve_from_file, write_to_file
 
+from algokit.cli.common.utils import MutuallyExclusiveOption
 from algokit.cli.tasks.utils import get_account_with_private_key
 
-
-def get_transaction(file: Path | None, transaction: str | None) -> Transaction:
-    if file:
-        txns: list[Transaction] = retrieve_from_file(str(file))  # type: ignore[no-untyped-call]
-        if len(txns) > 1:
-            raise click.ClickException("Only one transaction per file is supported.")
-        return txns[0]
-    elif transaction:
-        return cast(Transaction, encoding.msgpack_decode(transaction))  # type: ignore[no-untyped-call]
-    else:
-        raise click.ClickException("Provide either a file to sign or a transaction to sign, not both or none.")
+logger = logging.getLogger(__name__)
 
 
-def confirm_transaction(txn: Transaction) -> bool:
-    click.echo(json.dumps(txn.__dict__, indent=2))
-    response = click.prompt("Do you want to sign the above transaction?", type=click.Choice(["y", "n"]), default="n")
+def _format_transaction_for_stdout(txn: Transaction) -> dict[str, int | str | bytes | None]:
+    raw_txn = txn.__dict__.copy()
+
+    # Group isn't decoded via dictify by default, hence the need to decode it manually for stdout confirmation message.
+    if raw_txn.get("group"):
+        raw_txn["group"] = base64.b64encode(raw_txn["group"]).decode()
+
+    return raw_txn
+
+
+def _validate_for_signed_txns(txns: list[Transaction]) -> None:
+    signed_txns = [txn for txn in txns if isinstance(txn, SignedTransaction)]
+
+    if signed_txns:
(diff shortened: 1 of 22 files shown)


src/algokit/cli/tasks/sign_transaction.py
```diff
@@ -0,0 +1,63 @@
...
+    signed_txn = txn.sign(private_key)  # type: ignore[no-untyped-call]
+
+    if output:
+        write_to_file([signed_txn], str(output))  # type: ignore[no-untyped-call]
+        click.echo(f"Signed transaction written to {output}")
+    else:
+        click.echo(encoding.msgpack_encode({"txn": signed_txn.dictify()}))  # type: ignore[no-untyped-call]
+
+
+@click.command(name="sign", help="Sign an Algorand transaction.")
```
aorumbayev, line 40: @robdmoore just fyi, id suggest to use this as a starting point for pr review, remaining tweaks in other tasks are just minor polishing

src/algokit/cli/tasks/sign_transaction.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def get_transaction(file: Path | None, transaction: str | None) -> Transaction:
+    if file:
+        txns: list[Transaction] = retrieve_from_file(str(file))  # type: ignore[no-untyped-call]
+        if len(txns) > 1:
+            raise click.ClickException("Only one transaction per file is supported.")
+        return txns[0]
+    elif transaction:
+        return cast(Transaction, encoding.msgpack_decode(transaction))  # type: ignore[no-untyped-call]
```
robdmoore, line 19: Does this take base 64?

robdmoore: I more meant `transaction` itself should be a base64 string, is that the right input for that `msgpack_decode` method?

robdmoore: Also, if we use cast do we need to do some validation to check it is intact? Does it matter?

src/algokit/cli/tasks/sign_transaction.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def sign_and_output_transaction(txn: Transaction, private_key: str, output: Path | None) -> None:
+    signed_txn = txn.sign(private_key)  # type: ignore[no-untyped-call]
+
+    if output:
+        write_to_file([signed_txn], str(output))  # type: ignore[no-untyped-call]
+        click.echo(f"Signed transaction written to {output}")
+    else:
+        click.echo(encoding.msgpack_encode({"txn": signed_txn.dictify()}))  # type: ignore[no-untyped-call]
```
robdmoore, line 37: Is this right?

I thought the format of signed was `{txn: txn, sig: sig}`

src/algokit/cli/tasks/sign_transaction.py
```diff
@@ -0,0 +1,63 @@
...
+    help="The file to sign.",
+    required=False,
+)
+@click.option("--transaction", "-t", type=str, help="The transaction to sign.", required=False)
+@click.option("--output", "-o", type=Path, help="The output file.", required=False)
+@click.option("--force", is_flag=True, help="Force signing without confirmation.", required=False)
+def sign(*, account: str, file: Path | None, transaction: str | None, output: Path | None, force: bool) -> None:
+    signer_account = get_account_with_private_key(account)
+
+    if bool(file) == bool(transaction):
```
robdmoore, line 55: If both are false this won't trigger but should?

src/algokit/cli/tasks/sign_transaction.py
aorumbayev: all unsigned transaction objects should be inheriting from Transaction parent class. In case its already signed there is a separate validator that will throw an exception if user provided something that is signed already.

src/algokit/cli/tasks/sign_transaction.py
aorumbayev: @robdmoore added a custom click subclass for handling mutually exclusive options, this is not simplified a bit please check again

src/algokit/cli/tasks/sign_transaction.py
aorumbayev: @robdmoore as discussed during the call, i will assume goal for now, we can add extra flags to handle different formats in future - semantics of the commands itself should remain intact anyway

robdmoore (approved):

PR #326, opened 2023-10-11
feat: task send transaction(s)

aorumbayev:
# Proposed changes

- Adding ability to send transactions from goal clerk compatible signed txn objects
- Supports inputs via --file, --transaction options OR piped input

## Extra fixes

- Adding custom json encoder for sign method when confirmation popup shows up to display canonical goal clerk compatible txn fields AND fix the bug that would crash on encoding of AppCall transactions


src/algokit/cli/tasks/send_transaction.py
```diff
@@ -0,0 +1,166 @@
...
+        file_content = json.loads(raw_file_content)
+    except json.JSONDecodeError as ex:
+        raise click.ClickException("Invalid piped transaction content!") from ex
+
+    # Check if the content is a list of dicts with the required fields
+    if not isinstance(file_content, list) or not all(_is_sign_task_output_txn(item) for item in file_content):
+        raise click.ClickException("Invalid piped transaction content!")
+
+    # Convert the content into SignedTransaction objects
+    return [encoding.msgpack_decode(item["content"]) for item in file_content]  # type: ignore[no-untyped-call]
```
robdmoore, line 56: what is `"content"`?

src/algokit/cli/tasks/send_transaction.py
```diff
@@ -0,0 +1,166 @@
...
+    Args:
+        network (str): The network to which the transactions will be sent.
+        txns (list[SignedTransaction]): A list of signed transactions to be sent.
+
+    Returns:
+        None: The function does not return any value.
+    """
+    algod_client = load_algod_client(network)
+
+    if any(txn.transaction.group for txn in txns):
```
robdmoore, line 109: If the transaction group doesn't match on all of the transactions, what would happen?

robdmoore (approved):

src/algokit/cli/tasks/send_transaction.py
robdmoore: Should we use `transaction` instead to be clearer? What format is it in?

aorumbayev: @robdmoore renamed into `transaction_id` its just a string value.

src/algokit/cli/tasks/send_transaction.py
aorumbayev: @robdmoore algosdk will handle that and throw `inconsistent group values` error during sending the transaction group

PR #327, opened 2023-10-12
feat: ipfs upload

aorumbayev:
## Proposed changes

- Web3 storage based ipfs upload functionality with progress indicator

### DIFF
--- a/src/algokit/__init__.py
+++ b/src/algokit/__init__.py
@@ -1,16 +1,6 @@
-import platform
 import sys
 
 # this isn't beautiful, but to avoid confusing user errors we need this check before we start importing our own modules
-if sys.version_info < (3, 10, 0):
-    print(  # noqa: T201
-        f"Unsupported CPython version: {platform.python_version()} detected.\n"
-        "The minimum version of Python supported is CPython 3.10.\n"
-        "If you need help installing then this is a good starting point: \n"
-        "https://www.python.org/about/gettingstarted/",
-        file=sys.stderr,
-    )
-    sys.exit(-1)
 
 try:
     from algokit.core.log_handlers import initialise_logging, uncaught_exception_logging_handler
--- a/src/algokit/__init__.py
+++ b/src/algokit/__init__.py
@@ -1,6 +1,16 @@
+import platform
 import sys
 
 # this isn't beautiful, but to avoid confusing user errors we need this check before we start importing our own modules
+if sys.version_info < (3, 10, 0):
+    print(  # noqa: T201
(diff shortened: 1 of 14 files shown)


src/algokit/__init__.py
```diff
@@ -1,16 +1,6 @@
-import platform
 import sys
 
 # this isn't beautiful, but to avoid confusing user errors we need this check before we start importing our own modules
-if sys.version_info < (3, 10, 0):
```
robdmoore, line 5: Why is this removed?

aorumbayev: `Version block is outdated for minimum Python versionRuff[UP036](https://beta.ruff.rs/docs/rules/outdated-version-block)` ruff behaviour in my IDE, i guess my global ruff version is higher than the one in venv. Will revert that change manually

src/algokit/cli/tasks/ipfs.py
```diff
@@ -0,0 +1,111 @@
...
+        raise click.ClickException("You are not logged in! Please login using `algokit ipfs login`.")
+
+    try:
+        with file_path.open("rb") as file:
+            total = file_path.stat().st_size
+
+            if total > MAX_FILE_SIZE:
+                raise click.ClickException("File size exceeds 100MB limit!")
+
+            with click.progressbar(length=total, label="Uploading file") as bar:  # type: ignore[var-annotated]
```
robdmoore, line 86: Nice!

src/algokit/cli/tasks/ipfs.py
```diff
@@ -0,0 +1,111 @@
...
+    type=click.STRING,
+    help="Human readable name for this upload, for use in file listings.",
+)
+def upload(file_path: Path, name: str | None) -> None:
+    web3_storage_api_key = get_web3_storage_api_key()
+    if not web3_storage_api_key:
+        raise click.ClickException("You are not logged in! Please login using `algokit ipfs login`.")
+
+    try:
+        with file_path.open("rb") as file:
```
robdmoore, line 80: Do we need to check it's a file and not a directory? Are we adding ability to upload a directory?

robdmoore (approved):

src/algokit/cli/tasks/ipfs.py
aorumbayev: @robdmoore click takes care of that as part of the input argument validation `click.Path(exists=True, dir_okay=False, resolve_path=True, path_type=Path)`

dir_okay = implies no dirs allowed, it has to be an existing file and it will pass an abs path as Pathlib Path object to the input. 

And yeah we aren't supporting folder uploads at the moment

PR #328, opened 2023-10-12
feat: nfd domains lookups

aorumbayev:
## Proposed Changes

  - NFDomains forward/reverse lookups via https://api-docs.nf.domains/ mainnet api


robdmoore (approved):

tests/tasks/test_nfd_lookup.test_nfd_lookup_by_address_success.approved.txt
```diff
@@ -0,0 +1,2 @@
+DEBUG: HTTP Request: GET https://api.nf.domains/nfd/lookup?address=AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA&view=thumbnail&allowUnverified=false "HTTP/1.1 200 OK"
+"dummy.algo"
```
robdmoore, line 2: If the address option is selected maybe it shouldn't be dumped as json, but just the value? Easier to pipe around etc. then

PR #329, opened 2023-10-16
feat: task `mint` (arc3 or arc19 compliant)

aorumbayev:
## Proposed Changes

  - The PR adds a new command 'mint' to the 'algokit' CLI. This command allows users to mint new fungible or non-fungible assets on Algorand. The PR also includes validation for the inputs and provides feedback to the user about the minting process. Additionally, a new dependency 'py-multiformats-cid' is added to the project.

## TODO 

- [x] Snapshot tests
- [x] Clarifying on open questions


aorumbayev:
@Loedn @robdmoore adding some open questions on semantics of the command:

1. Do we want to add extra arguments on allowing user to optionally set same creator address on `clawback` and `freeze` fields? By default im setting both to None  
2. Do we want to allow user to toggle the `default_frozen` boolean parameter on acfg txn paramaters? By default im using false 
3. Since arc19 is all about mutability, im not allowing user to set manager field to an empty string in case arc19 token is minted, in case of arc3 i do prompt for confirmation whether user needs this to be mutable or not. Does it make sense to allow user to set empty manager on supposedly mutable arc19 compliant assets ? ( i guess it won't be idiomatic arc19 in that case anymore?)

src/algokit/cli/tasks/mint.py
```diff
@@ -0,0 +1,163 @@
...
+from algokit.core.tasks.mint.models import TokenMetadata
+
+logger = logging.getLogger(__name__)
+
+
+def _validate_inputs(total: int, decimals: int) -> None:
+    # Validate total and decimals
+    if not (total == 1 or (total % 10 == 0 and total != 0)):
+        raise click.ClickException("Total must be 1 or a power of 10 larger than 1 (10, 100, 1000, ...).")
+    if not ((total == 1 and decimals == 0) or (total != 1 and decimals == int(math.log10(total)))):
```
robdmoore, line 24: This is true for a fractional NFT, but for a fungible token total and decimals can be whatever right?

src/algokit/cli/tasks/mint.py
```diff
@@ -0,0 +1,163 @@
...
+)
+@click.option(
+    "--creator",
+    "-r",
+    required=True,
+    prompt="Please enter the address or alias of the asset creator",
+    help="Address or alias of the asset creator.",
+    type=click.STRING,
+)
+@click.option(
```
robdmoore, line 43: Should we validate max length?

src/algokit/cli/tasks/mint.py
```diff
@@ -0,0 +1,163 @@
...
+    prompt="""Please enter the number of decimals. Set 0 for a pure NFT,
+        or equal to the logarithm in base 10 of total supply for a fractional NFT.""",
+    help="""Number of decimals. Set 0 for a pure NFT, ",
+        "or equal to the logarithm in base 10 of total supply for a fractional NFT.""",
+)
+@click.option(
+    "--image-path",
+    "image_path",
+    type=click.Path(exists=True, dir_okay=False, file_okay=True, resolve_path=True, path_type=Path),
+    prompt="Please enter the path to the ARC19 compliant asset metadata file to be uploaded to IPFS",
```
robdmoore, line 83: I'm not sure the prompt and help text are correct? They talk about metadata rather than image.

src/algokit/cli/tasks/mint.py
```diff
@@ -0,0 +1,163 @@
...
+    "--token-metadata-path",
+    "token_metadata_path",
+    type=click.Path(exists=True, dir_okay=False, file_okay=True, resolve_path=True, path_type=Path),
+    help="""Path to the ARC19 compliant asset metadata file to be uploaded to IPFS. If not provided,
+        a default metadata object will be generated automatically based on asset-name, decimals and image.
+        For more details refer to https://arc.algorand.foundation/ARCs/arc-0003#json-metadata-file-schema.""",
+    default=None,
+    required=False,
+)
+@click.option(
```
robdmoore, line 97: Given we are making things simple for people, I wonder if this question should instead be about mutability rather than metadata standard? Hypothesis: someone that understands the metadata standards probably can write the code?

src/algokit/core/tasks/mint/mint.py
```diff
@@ -0,0 +1,154 @@
...
+    metadata_cid = upload_to_web3_storage(
+        token_metadata.to_file_path(),
+        api_key=api_key,
+    )
+    logger.info(f"Metadata uploaded to Web3 Storage: {metadata_cid}")
+
+    asset_config_params = AssetConfigTxnParams(
+        sender=creator_account.address,
+        sp=client.suggested_params(),
+        reserve=_reserve_address_from_cid(metadata_cid),
```
robdmoore, line 134: This should only happen for ARC-19?

robdmoore (approved):

src/algokit/cli/tasks/mint.py
aorumbayev: CleanShot 2023-10-16 at 19 12 41
I primarily based this of this chunk in https://arc.algorand.foundation/ARCs/arc-0003, assumed this implies that ARC3 spec is only for NFTs (fractional or not)

src/algokit/core/tasks/mint/mint.py
aorumbayev: @robdmoore good catch, should i use creator_address by default for arc3 then or keep it None?

src/algokit/cli/tasks/mint.py
aorumbayev: @robdmoore added validators

src/algokit/core/tasks/mint/mint.py
aorumbayev: @robdmoore setting it to None for case of arc3

src/algokit/cli/tasks/mint.py
aorumbayev: @robdmoore renamed to `mutable`

src/algokit/cli/tasks/mint.py
aorumbayev: @robdmoore added extra flag --nft/--ft for checking when above validators need to be checked, if --ft is set then anything in total/decimals is accepted

robdmoore (approved):

PR #336, opened 2023-10-19
fix: docker compose ps parsing for version >= 2.21

### LINKED ISSUE #335: docker compose 2.21.0 introduces breaking change to `ps` JSON output
See https://github.com/docker/compose/pull/10918

This version of docker was released 8/20/2023 so anyone with a recent version of docker compose will run into an exception when trying to run any `algokit localnet`command. 

The fix should be pretty easy, so I'll PR shortly.

joe-p:
Fixes #335 

## Proposed Changes

  - Support both the new and old `docker compose ps --format json` formats

### DIFF
--- a/src/algokit/core/sandbox.py
+++ b/src/algokit/core/sandbox.py
@@ -116,7 +116,21 @@ class ComposeSandbox:
         )
         if run_results.exit_code != 0:
             return []
-        data = json.loads(run_results.output)
+
+        data = []
+
+        # docker compose version >= 2.21.0 outputs a blank line when nothing is running
+        # json.loads with a blank line will throw an error
+        if len(run_results.output.strip()) > 0:
+            # docker compose version < 2.21.0 outputs a JSON array, so it will have no newlines
+            if run_results.output.strip().find("\n") == -1:
+                data = json.loads(run_results.output)
+            # docker compose version >= 2.21.0 outputs seperate JSON objects, each on a new line
+            else:
+                for line in run_results.output.split("\n"):
+                    if len(line) > 0:
+                        data.append(json.loads(line))
+
         assert isinstance(data, list)
         return cast(list[dict[str, Any]], data)
 


src/algokit/core/sandbox.py
```diff
@@ -116,7 +116,21 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
         )
         if run_results.exit_code != 0:
             return []
-        data = json.loads(run_results.output)
+
+        data = []
+
+        # docker compose version >= 2.21.0 outputs a blank line when nothing is running
+        # json.loads with a blank line will throw an error
+        if len(run_results.output.strip()) > 0:
```
achidlow, line 124: `if len(x) > 0:` is more idiomatic as just `if x:`

src/algokit/core/sandbox.py
```diff
@@ -116,7 +116,21 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
...
+            # docker compose version < 2.21.0 outputs a JSON array, so it will have no newlines
+            if run_results.output.strip().find("\n") == -1:
+                data = json.loads(run_results.output)
+            # docker compose version >= 2.21.0 outputs seperate JSON objects, each on a new line
+            else:
+                for line in run_results.output.split("\n"):
```
achidlow, line 130: `split("\n")` is more idiomatic as `splitlines()`

src/algokit/core/sandbox.py
achidlow: also, I would assign the result of strip. you could even use walrus operator for that:
`if stripped_output := run_results.output.strip():`

src/algokit/core/sandbox.py
achidlow: also, could you use a comprehension here? 
`data = [json.loads(line) for line in stripped_output.splitlines() if line]` ?

achidlow: (also, could you omit the `if line` since you can run on `stripped_output` instead?)

joe-p:
@achidlow I realized I could simplify the condition since the old output always starts with "[" and the new output doesn't. Used the suggested list comprehension as well.

aorumbayev (approved):

src/algokit/core/sandbox.py
```diff
@@ -116,7 +116,16 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
         )
         if run_results.exit_code != 0:
             return []
-        data = json.loads(run_results.output)
+
+        data = []
+
+        # `docker compose ps --format json` on version < 2.21.0 outputs a JSON arary
+        if run_results.output and run_results.output[0] == "[":
```
achidlow, line 123:
```suggestion
        if run_results.output.startswith("["):
```

src/algokit/core/sandbox.py
```diff
@@ -116,7 +116,16 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
         )
         if run_results.exit_code != 0:
             return []
-        data = json.loads(run_results.output)
+
+        data = []
```
achidlow, line 120: can this line be removed now? if that impacts the type checker, might just need to explicitly annotate the assignment in the if branch

joe-p: Removed in 9709b8564bc1f4bcc0f958e58fba452745f9ff23

src/algokit/core/sandbox.py
```diff
@@ -116,7 +116,14 @@ def ps(self, service_name: str | None = None) -> list[dict[str, Any]]:
...
         if run_results.exit_code != 0:
             return []
-        data = json.loads(run_results.output)
+
+        # `docker compose ps --format json` on version < 2.21.0 outputs a JSON arary
+        if run_results.output.startswith("["):
+            data = json.loads(run_results.output)
+        # `docker compose ps --format json` on version >= 2.21.0 outputs seperate JSON objects, each on a new line
+        else:
+            data = [json.loads(line) for line in run_results.output if line]
```
aorumbayev, line 125:
```suggestion
            data = [json.loads(line) for line in run_results.output.splitlines() if line]
```

aorumbayev: @joe-p  Just double checked on mac, produces `Unhandled JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)` unless splitlines are applied. Otherwise lgtm - will merge as soon as you commit suggestion, and thanks for tackling this 👍

aorumbayev (changes requested):

PR #346, opened 2023-11-09
fix: support the multiple file outputs of goal clerk split

neilcampbell:
## Fixes to goal command
  - Add support for copying the multiple output files that goal clerk split generates
  - Add support for file paths like `./my-file.ext` and `~/my-file.ext`
  - Support additional output arguments available in the goal commands

### DIFF
--- a/src/algokit/core/goal.py
+++ b/src/algokit/core/goal.py
@@ -41,40 +41,56 @@ def list_files_in_volume(volume_path: Path) -> list[str]:
 
 def preprocess_command_args(
     command: list[str], volume_mount_path_local: Path, docker_mount_path_local: Path
-) -> tuple[list[str], list[Path], list[str]]:
-    input_filenames = []
-    output_filenames = []
+) -> tuple[list[Path], list[Path], list[str]]:
+    input_files = []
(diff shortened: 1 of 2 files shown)


src/algokit/core/goal.py
```diff
@@ -41,40 +41,56 @@ def list_files_in_volume(volume_path: Path) -> list[str]:
...
     try:
         for i, arg in enumerate(command):
             if is_path_or_filename(arg):
-                arg_path = Path(arg)
-                arg_changed = docker_mount_path_local.joinpath(arg_path.name)
+                absolute_arg_path = Path(arg).expanduser().absolute()
+                arg_changed = docker_mount_path_local.joinpath(absolute_arg_path.name)
                 command[i] = str(arg_changed)
 
-                file_exists = arg_path.exists() or Path.cwd().joinpath(arg_path.name).exists()
```
neilcampbell, line 54: @negar-abbasi Do you remember why the `Path.cwd().joinpath(arg_path.name).exists()` part was here? It doesn't appear to be needed from what I could tell.

robdmoore (approved):

PR #348, opened 2023-11-10
Support codespaces for explore

joe-p:
Adds support for GitHub codespaces with algokit explore, similar to how gitpod was supported.

Future work to be done on bootstrap so that it will setup the .env to the proper urls (opposed to localhost) for the frontend.

### DIFF
--- a/src/algokit/cli/explore.py
+++ b/src/algokit/cli/explore.py
@@ -28,7 +28,7 @@ class NetworkConfiguration(NetworkConfigurationRequired, total=False):
 
 
 GITPOD_URL = os.environ.get("GITPOD_WORKSPACE_URL")
-CODESPAE_NAME = os.environ.get("CODESPACE_NAME")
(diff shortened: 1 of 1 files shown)


src/algokit/cli/explore.py
```diff
@@ -28,17 +28,41 @@ class NetworkConfiguration(NetworkConfigurationRequired, total=False):
 
 
 GITPOD_URL = os.environ.get("GITPOD_WORKSPACE_URL")
+CODESPAE_NAME = os.environ.get("CODESPACE_NAME")
```
robdmoore, line 31: Misspelt

robdmoore (approved):

src/algokit/cli/explore.py
joe-p: Fixed in b5ce9a2539d5edba2c0fd1ea8188372ce807f3c1

PR #351, opened 2023-11-13
feat: migrating localnet to latest indexer v3.x images

aorumbayev:
## Proposed Changes

  - Adds support and migrates to -> official indexer v3 image (latest tag as of today)
  - Adds required dependency on conduit which forces using extra algod follower container that does slow down execution of certain operations as well as boot up and shut down time of compose environment

### Related improvements

- Adding pytest-xdist to speed up test execution

### DIFF
--- a/.github/workflows/build-python.yaml
+++ b/.github/workflows/build-python.yaml
@@ -31,7 +31,7 @@ jobs:
         shell: bash
         run: |
           set -o pipefail
-          poetry run pytest --junitxml=pytest-junit.xml --cov-report=term-missing:skip-covered --cov=src | tee pytest-coverage.txt
+          poetry run pytest -n auto --junitxml=pytest-junit.xml --cov-report=term-missing:skip-covered --cov=src | tee pytest-coverage.txt
 
       - name: pytest coverage comment - using Python 3.10 on ubuntu-latest
         if: matrix.python == '3.10' && matrix.os == 'ubuntu-latest'
--- a/poetry.lock
+++ b/poetry.lock
@@ -897,6 +897,20 @@ files = [
 test = ["pytest (>=6)"]
 
 [[package]]
+name = "execnet"
+version = "2.0.2"
+description = "execnet: rapid multi-Python deployment"
+optional = false
+python-versions = ">=3.7"
+files = [
+    {file = "execnet-2.0.2-py3-none-any.whl", hash = "sha256:88256416ae766bc9e8895c76a87928c0012183da3cc4fc18016e6f050e025f41"},
+    {file = "execnet-2.0.2.tar.gz", hash = "sha256:cc59bc4423742fd71ad227122eb0dd44db51efb3dc4095b45ac9a08c770096af"},
+]
+
+[package.extras]
+testing = ["hatch", "pre-commit", "pytest", "tox"]
+
+[[package]]
 name = "filelock"
 version = "3.12.2"
 description = "A platform independent file lock."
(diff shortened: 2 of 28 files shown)


aorumbayev:
@winder a few further questions to clarify if possible (cc @robdmoore @neilcampbell )

I was able to execute the algod container with 2 networks (relay participation and one follower nodes). It wasn't clear that network configurations are passed differently that config.json which is why i was confused initially. If you execute the compose on the contents of the attached archive you will see that it seems to be working fine. Conduit is able to chat with 8081 port on algod (which is the port for the follower, while relay is at 8080). There is a slight millisecond delay that still happens with this setup where fresh deployment of an app takes certain amount of time before it pops up on indexer search application by account endpoint response -> **will require adding some tweaks to our fixtures on utils ts and py** (cc @neilcampbell @robdmoore ) 

1. The configuration attached is generated from the compose file [here](https://github.com/algorandfoundation/algokit-cli/blob/4dc473e02a049b615410f756a4ead37acdc27c87/src/algokit/core/sandbox.py#L354) - for the sake of simplicity for yourself to check it out. Despite setup working as expected (conduit being able to chat with second node within single algod container and it running with DEV_MODE env var), i am seeing the following error logs reported by follower:
```
2023-11-15 00:35:48 {"Context":"sync","file":"service.go","function":"github.com/algorand/go-algorand/catchup.(*Service).fetchAndWrite","level":"info","line":312,"msg":"fetchAndWrite(14): Aborted while waiting for lookback block to ledger after failing once : wsFetcherClient(127.0.0.1:44360).requestBlock(14): Request failed: requested block is not available","name":"0.0.0.0:10000","time":"2023-11-14T23:35:48.635482Z"}
```
Which i assume implies that follower node is unable to fetch the data from relay on port 10000 ? The port 10000 is set as a `GOSSIP_PORT` which i set similarly to your gist at https://gist.github.com/winder/8e167d94817370c8ed2e8579432f9003. Removing this value completely fails on `algokit_algod              | Error starting deployed network: open /algod/data/algod-listen.net: no such file or directory`. What can be root cause for these errors or its just a warning that can be ignored in localnet scenarios (assuming that it seemingly doesn't affect anything)?

2. Do i need to specify `GENESIS_ADDRESS` and `PEER_ADDRESS` env vars on an algod container if both follower and relay are running in a single container? In the gist setup sample https://gist.github.com/winder/8e167d94817370c8ed2e8579432f9003 those are passed explicitly to the follower container but i am unsure whether i need to set them or not in the case of having follower inside first container. Perhaps this is also somehow causing the errors from question 1?
[Archive.zip](https://github.com/algorandfoundation/algokit-cli/files/13358320/Archive.zip)


If you need any extra context/input please let me know.

winder:
What behavior do you see that is undesirable?

The log you shared is an info level event, so not an error.

aorumbayev:
@winder aside from compose working the only side effect observed was on our algokit utils unit tests in few places where we get to test app update calls we perform 2 consecutive deploys of a contract. In the case with current v2 indexer, changes are propagated almost instantly so its possible to perform `lookup_account_application_by_creator` right after app is created (expecting that it will find a freshly minted app id). In case of running against this compose there is still a slight millisecond delay. But i assume its expected behaviour given that there are is an extra hop for data with extra follower and conduit as middleware. If you can confirm whether its reasonable to expect very short delays for data to propagate from algod to indexer in such localnet scenarios, even when follower node is within the same container then all good. 

Other than that would still appreciate a reply for question 2. Just want to ensure i'm not passing any unnecessary environment variables or configurations that can cause unforeseen side effects. 

After that will publish the pr and we will do more extensive smoke tests + will require changes in unit tests to account for increased delays on certain tests (that were previously not accounting for the fact that it may take extra time for indexer to sync with changes from algod). CC @robdmoore @neilcampbell

robdmoore (approved):

neilcampbell (approved):

robdmoore (approved):

Issue #356, opened 2023-11-14
Update Python VSCode settings


Loedn:
The old settings for formatting/linting have all been deprecated in favour of dedicated extensions. We need to update all of our repositories to reflect such upgrade.

Useful docs vscode-python [release notes](https://github.com/microsoft/vscode-python/releases/tag/v2023.20.0), migration [guide](https://github.com/microsoft/vscode-python/wiki/Migration-to-Python-Tools-Extensions).

We should also explore using Ruff only rather than Ruff+Black as Ruff is faster

negar-abbasi:
The repos are now update with the new VSCode-Python formatting/linting settings.
While Ruff formatter shows promise for faster processing, it's currently in a beta yet production-ready state. Considering this, we've decided to stick with the Black formatter for our templates until there is a stable release of the Ruff formatter. However, for repos like `algokit-cli` and `algokit-utils-py`, where we can swiftly manage changes, we've switched to Ruff for formatting.

PR #359, opened 2023-11-21
chore: ruff improvements

negar-abbasi:
Fixes #

## Proposed Changes

  - updating vscode setting as formatting/linting have been deprecated in favour of dedicated extensions
  - working with ruff formatter instead of black

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,7 +3,8 @@
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
     // Don't want to use isort because it conflicts with Ruff - see run on save below
-    "source.organizeImports": false
+    "source.organizeImports": false,
+    "source.fixAll": true
   },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "files.exclude": {
@@ -32,33 +33,34 @@
     }
   },
   "python.analysis.extraPaths": ["${workspaceFolder}/src"],
-  "python.formatting.provider": "black",
   "[python]": {
     // https://dev.to/eegli/quick-guide-to-python-formatting-in-vs-code-2040
-    "editor.defaultFormatter": null
+    "editor.defaultFormatter": "charliermarsh.ruff"
   },
   "python.analysis.typeCheckingMode": "basic",
-  "python.linting.enabled": true,
-  "python.linting.lintOnSave": true,
-  "ruff.importStrategy": "fromEnvironment",
-  "python.linting.pylintEnabled": false,
-  "python.linting.mypyEnabled": false,
+  "ruff.lint.run": "onSave",
+  "ruff.lint.args": ["--config=pyproject.toml"], // not sure yet
+  "ruff.importStrategy": "fromEnvironment", // needed
+  "ruff.fixAll": true, //lint and fix all files in workspace   // not needed
+  "ruff.organizeImports": false, //organize imports on save   // not needed
+
   "mypy.configFile": "pyproject.toml",
   // set to empty array to use config from project
   "mypy.targets": [],
   "mypy.runUsingActiveInterpreter": true,
-  "python.linting.banditEnabled": false,
-  "python.linting.prospectorEnabled": false,
-  "python.linting.pydocstyleEnabled": false,
-  "python.linting.pycodestyleEnabled": false,
   "python.testing.unittestEnabled": false,
   "python.testing.pytestEnabled": true,
   "emeraldwalk.runonsave": {
     "commands": [
+      // Run Ruff formatter on save of Python file
+      // {
+      //   "match": "\\.py$",
+      //   "cmd": "${workspaceFolder}/.venv/bin/ruff format ${file}"
+      // },
       // Run Ruff linter on save of Python file
       {
         "match": "\\.py$",
-        "cmd": "${workspaceFolder}/.venv/bin/ruff ${file} --fix"
+        "cmd": "${workspaceFolder}/.venv/bin/ruff ${file} --fix" // not sure it is really needed -- see without extension
       }
     ]
   },
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -38,11 +38,18 @@
     "editor.defaultFormatter": "charliermarsh.ruff"
   },
   "python.analysis.typeCheckingMode": "basic",
+  "ruff.enable": true,
   "ruff.lint.run": "onSave",
-  "ruff.lint.args": ["--config=pyproject.toml"], // not sure yet
-  "ruff.importStrategy": "fromEnvironment", // needed
-  "ruff.fixAll": true, //lint and fix all files in workspace   // not needed
-  "ruff.organizeImports": false, //organize imports on save   // not needed
+  "ruff.lint.args": ["--config=pyproject.toml"],
+  "ruff.importStrategy": "fromEnvironment",
+  "ruff.fixAll": true, //lint and fix all files in workspace
+  "ruff.organizeImports": true, //organize imports on save
+  "ruff.codeAction.disableRuleComment": {
+    "enable": true
+  },
+  "ruff.codeAction.fixViolation": {
+    "enable": true
+  },
 
   "mypy.configFile": "pyproject.toml",
   // set to empty array to use config from project
@@ -52,12 +59,6 @@
   "python.testing.pytestEnabled": true,
   "emeraldwalk.runonsave": {
     "commands": [
-      // Run Ruff formatter on save of Python file
-      // {
-      //   "match": "\\.py$",
-      //   "cmd": "${workspaceFolder}/.venv/bin/ruff format ${file}"
-      // },
-      // Run Ruff linter on save of Python file
       {
         "match": "\\.py$",
         "cmd": "${workspaceFolder}/.venv/bin/ruff ${file} --fix" // not sure it is really needed -- see without extension
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,7 +3,7 @@
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
     // Don't want to use isort because it conflicts with Ruff - see run on save below
-    "source.organizeImports": false,
+    "source.organizeImports": true,
     "source.fixAll": true
   },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -34,7 +34,6 @@
   },
   "python.analysis.extraPaths": ["${workspaceFolder}/src"],
   "[python]": {
-    // https://dev.to/eegli/quick-guide-to-python-formatting-in-vs-code-2040
     "editor.defaultFormatter": "charliermarsh.ruff"
   },
   "python.analysis.typeCheckingMode": "basic",
@@ -61,7 +60,7 @@
     "commands": [
       {
         "match": "\\.py$",
-        "cmd": "${workspaceFolder}/.venv/bin/ruff ${file} --fix" // not sure it is really needed -- see without extension
+        "cmd": "${workspaceFolder}/.venv/bin/ruff ${file} --fix"
       }
     ]
   },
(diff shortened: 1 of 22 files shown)


.vscode/settings.json
```diff
@@ -3,7 +3,8 @@
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
     // Don't want to use isort because it conflicts with Ruff - see run on save below
-    "source.organizeImports": false
+    "source.organizeImports": false,
```
robdmoore, line 6: We might be able to make this true if isort doesn't get run anymore with the default formatter change

pyproject.toml
```diff
@@ -115,6 +114,10 @@ ignore = [
   "ANN102", # no type for cls
   "SIM108", # allow if-else in place of ternary
   "RET505", # allow else after return
+  "Q001", # bad quotes
+  "Q002", # bad quotes docstring
+  "Q003", # avoidable escaped quotes
```
robdmoore, line 119: Are the ruff settings compatible with Black formatting we had previously?

pyproject.toml
```diff
@@ -143,9 +146,6 @@ target-version = "py310"
 allow-star-arg-any = true
 suppress-none-returning = true
 
-[tool.black]
-line-length = 120
```
robdmoore, line 147: Is this the default for ruff or should we specify it?

src/algokit/cli/tasks/mint.py
```diff
@@ -211,7 +211,10 @@ def mint(  # noqa: PLR0913
 
     client = load_algod_client(network)
     validate_balance(
-        client, creator_account, 0, algos_to_microalgos(ASSET_MINTING_MBR)  # type: ignore[no-untyped-call]
+        client,
```
robdmoore, line 214: Is this a formatting change? Ideally we want to avoid formatting differences to reduce churn on source files as they get edited.

robdmoore (approved):

pyproject.toml
negar-abbasi: The default is 88. It has been specified [here](https://github.com/algorandfoundation/algokit-cli/blob/472e6799546a5439c42c7ffedc30620096638c34/pyproject.toml#L61)

src/algokit/cli/tasks/mint.py
negar-abbasi: Ruff claims that Black-formatted projects like Django and Zulip, > 99.9% of lines are formatted identically and non-Black-formatted code, the formatter makes some different decisions than Black.
So either way there will be little changes. I think this is one of the 0.01% cases.

pyproject.toml
negar-abbasi: Yes for this project they are ok. I will add more of the conflicts for the later as well.

src/algokit/cli/tasks/mint.py
achidlow: https://docs.astral.sh/ruff/formatter/black/

CONTRIBUTING.md
```diff
@@ -16,7 +16,7 @@ AlgoKit development is done within the [AlgoKit Guiding Principles](./docs/algok
 2. Install pre-requisites:
 
    - Manually:
-     - Install `Python` - [Link](https://www.python.org/downloads/): The minimum required version is `3.10`. It is also recommended to use `3.10` for development so dependencies on higher versions are not introduced, and to avoid debugging issues with `3.11`(see https://github.com/fabioz/PyDev.Debugger/issues/234 and https://github.com/microsoft/debugpy/issues/939)
+     - Install `Python` - [Link](https://www.python.org/downloads/): The minimum required version is `3.10`. It is also recommended to use `3.10` for development so dependencies on higher versions are not introduced, and to avoid debugging issues with `3.10`(see <https://github.com/fabioz/PyDev.Debugger/issues/234> and <https://github.com/microsoft/debugpy/issues/939>)
```
achidlow, line 19: this should still say `3.11`

pyproject.toml
```diff
@@ -113,8 +112,15 @@ select = [
...
-  "SIM108", # allow if-else in place of ternary
+  "E111", # indentation is not a multiple of four
+  "E117", # over-indented
+  "Q000", # bad quotes inline string
+  "Q001", # bad quotes multiline string
+  "Q002", # bad quotes docstring
+  "Q003", # avoidable escaped quotes
   "RET505", # allow else after return
+  "SIM108", # allow if-else in place of ternary
+  "W191", # indentation contains tabs
```
achidlow, line 123: indentation should never contain tabs

pyproject.toml
achidlow: why did these need to be added? in general, trying to keep a minimal ignore list is preferred

src/algokit/core/proc.py
```diff
@@ -86,7 +86,7 @@ def run_interactive(
     command_str = " ".join(command)
     logger.debug(f"Running '{command_str}' in '{cwd or Path.cwd()}'")
 
-    result = subprocess_run(command, cwd=cwd, env=env)
+    result = subprocess_run(command, cwd=cwd, env=env, check=True)
```
achidlow, line 89: this should't be `check=True` that will cause an exception, the result code is being checked explicitly below instead.
did it hit a linting rule? if so, can explicitly specify default of `check=False`

src/algokit/core/tasks/vanity_address.py
```diff
@@ -124,7 +124,7 @@ def generate_vanity_address(keyword: str, match: MatchType) -> VanityAccount:
     """
     jobs: list[Process] = []
 
-    def signal_handler(sig: int, frame: types.FrameType | None) -> Any | None:
+    def signal_handler(sig: int, frame: types.FrameType | None) -> Any | None:  # noqa: ANN401
```
achidlow, line 127: you could type this as returning `object` instead I think? then no need for the `noqa`

src/algokit/core/version_prompt.py
```diff
@@ -44,7 +44,7 @@ def get_latest_version_or_cached() -> str | None:
     version_check_path = get_app_state_dir() / "last-version-check"
 
     try:
-        last_checked = os.path.getmtime(version_check_path)
+        last_checked = Path.stat(version_check_path).st_mtime
```
achidlow, line 47: since `version_check_path` is already a `Path` object, this should just be `version_check_path.stat().st_mtime`

tests/init/test_init.py
```diff
@@ -253,7 +253,7 @@ def test_init_minimal_interaction_required_yes_git_no_network(
     paths = {p.relative_to(created_dir) for p in created_dir.iterdir()}
     assert paths == {Path(".git"), Path("myapp")}
     git_rev_list = subprocess.run(
-        ["git", "rev-list", "--max-parents=0", "HEAD"], cwd=created_dir, capture_output=True, text=True
+        ["git", "rev-list", "--max-parents=0", "HEAD"], cwd=created_dir, capture_output=True, text=True, check=True
```
achidlow, line 256: similarly, the `returncode` here is being asserted below, which is prefereable to raising an exception here (ie causes a test failure rather than a test error)

tests/init/test_init.py
```diff
@@ -456,7 +456,7 @@ def test_init_ask_about_git(tmp_path_factory: TempPathFactory, mock_questionary_
     paths = {p.relative_to(created_dir) for p in created_dir.iterdir()}
     assert paths == {Path("myapp"), Path(".git")}
     git_rev_list = subprocess.run(
-        ["git", "rev-list", "--max-parents=0", "HEAD"], cwd=created_dir, capture_output=True, text=True
+        ["git", "rev-list", "--max-parents=0", "HEAD"], cwd=created_dir, capture_output=True, text=True, check=True
```
achidlow, line 459: ditto

src/algokit/core/proc.py
negar-abbasi: It did, cool. I will set to False.

tests/localnet/conftest.py
```diff
@@ -7,13 +7,13 @@
 
 
 @pytest.fixture(autouse=True)
-def algod_health_fast_timings(mocker: MockerFixture) -> None:  # noqa: ignore[PT004]
+def algod_health_fast_timings(mocker: MockerFixture) -> None:  # noqa: PT004: ignore[PT004]
```
achidlow, line 10: why were we suppressing these instead of renaming the function to meet the linter rule?

tests/localnet/conftest.py
```diff
@@ -7,13 +7,13 @@
...
     mocker.patch("algokit.core.sandbox.DEFAULT_WAIT_FOR_ALGOD", 0.1)
     mocker.patch("algokit.core.sandbox.DEFAULT_HEALTH_TIMEOUT", 0.1)
 
 
 @pytest.fixture()
-def health_success(httpx_mock: HTTPXMock) -> None:  # noqa: ignore[PT004]
+def health_success(httpx_mock: HTTPXMock) -> None:  # noqa: PT004: ignore[PT004]
```
achidlow, line 16: ditto

pyproject.toml
negar-abbasi: We are adding their parent categories to be checked (like Q or E, W)

src/algokit/cli/tasks/mint.py
robdmoore: Cool, happy!

pyproject.toml
achidlow: oh cool, maybe group those together and have a comment around them pointing to that doc?

src/algokit/__init__.py
```diff
@@ -2,7 +2,7 @@
 import sys
 
 # this isn't beautiful, but to avoid confusing user errors we need this check before we start importing our own modules
-if sys.version_info < (3, 10, 0):
+if sys.version_info < (3, 10, 0):  # noqa: UP036
```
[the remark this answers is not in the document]
neilcampbell: Yeah it's weird, but reading https://github.com/astral-sh/ruff/issues/7258 what were doing by setting `target-version` is telling Ruff that we don't support versions less than that, so it's ok to suggest that you remove that code block (which is the suggested fix for UP036).

src/algokit/core/proc.py
neilcampbell: @negar-abbasi This seems to be still be `check=True`

debug.py
```diff
@@ -47,7 +47,7 @@
 os.chdir(vcs_root / "src")
 for command in commands_sequence or [""]:
     click.echo(click.style(f"> algokit -v {command}", bold=True), err=True)
-    run_result = subprocess.run([sys.executable, "-m", "algokit", "-v", *command.split()])
+    run_result = subprocess.run([sys.executable, "-m", "algokit", "-v", *command.split()], check=True)
```
neilcampbell, line 50: Should this be `check=False` as well?

neilcampbell (approved):

robdmoore:
Closes #356

negar-abbasi:
@robdmoore, in updating our templates, we got a warning regarding the ruff format:
> warning: `ruff format` is a work-in-progress, subject to change at any time, and intended for internal use only.
reviewing this with @neilcampbell, we realized that [ruff is in a beta version but production ready state](https://astral.sh/blog/the-ruff-formatter#production-ready). Given this uncertainty, it seems better to continue using the Black formatter for now and hold off on integrating ruff into our templates until it reaches a more stable release phase.

neilcampbell:
Just to add a bit more context to the above, we are planning to switch to Ruff for formatting in the repos we control and can easily handle any breaking changes in. We will however hold off on updating the templates until there is a stable release of the Ruff formatter, as to minimise any potentially impact to developer projects created from the templates.

achidlow:
Yep, that's a good plan - that's what I thought we'd be doing :)

.vscode/settings.json
```diff
@@ -3,7 +3,8 @@
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
     // Don't want to use isort because it conflicts with Ruff - see run on save below
-    "source.organizeImports": false
+    "source.organizeImports": true,
```
neilcampbell, line 6: You can probably remove the comment now.

PR #370, opened 2023-12-14
feat: adding algokit analyze - perform static analysis with tealer integration

aorumbayev:
## Proposed Changes

  - Initial implementation of algokit analyze feature


aorumbayev:
@neilcampbell i opted for not using `prettytable` in the end. This is more readable and size-friendly. Table will look bad on terminals with small width given that we gotta pass large urls and etc

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+            pattern = "**/*.teal" if recursive else "*.teal"
+            input_files.extend(sorted(input_path.glob(pattern)))
+        else:
+            input_files.append(input_path)
+    return sorted(set(input_files))
+
+
+@click.command(
+    name="analyze",
+    help="Analyze TEAL programs for common vulnerabilities with AlgoKit Tealer integration.",
```
neilcampbell, line 105:
```suggestion
    help="Analyze TEAL programs for common vulnerabilities using Tealer.",
```

docs/cli/index.md
```diff
@@ -727,6 +736,44 @@ Collection of useful tasks to help you develop on Algorand.
...
+Analyze TEAL programs for common vulnerabilities with AlgoKit Tealer integration.
+
+```shell
+algokit task analyze [OPTIONS]
+```
+
+### Options
+
+
+### -f, --file <files>
```
neilcampbell, line 750: These docs need updating, based on the recent changes.

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+    nargs=-1,
+    type=click.Path(exists=True, dir_okay=True, file_okay=True, path_type=Path),
+    required=True,
+)
+@click.option(
+    "-r",
+    "--recursive",
+    is_flag=True,
+    cls=MutuallyExclusiveOption,
+    not_required_if=["file"],
```
neilcampbell, line 118: Does this work if a file and a directory is passed?

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+@click.command(
+    name="analyze",
+    help="Analyze TEAL programs for common vulnerabilities with AlgoKit Tealer integration.",
+)
+@click.argument(
+    "input_paths",
+    nargs=-1,
+    type=click.Path(exists=True, dir_okay=True, file_okay=True, path_type=Path),
```
neilcampbell, line 110: Assuming you could pass a file and a directory at the same time?

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+    is_flag=True,
+    help="Force verification without the disclaimer confirmation prompt.",
+)
+@click.option(
+    "-b",
+    "--baseline",
+    is_flag=True,
+    help=(
+        "Exit with a non-zero code if any diffs are identified between the current "
+        "and baseline reports. By default baseline reports are stored in the "
```
neilcampbell, line 132:
```suggestion
        "and baseline reports. By default reports are stored in the "
```

Assumption here is that reports are generated every time the command is run. The baseline flag is simply a mechanism for failing when execution doesn't match the last previously generated report.

Also wanted to confirm that a report is not written when the --baseline flag is specified and a diff is detected? This means we can run the exact command again and get the same result.

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+    ),
+)
+@click.option(
+    "-o",
+    "--output",
+    "output_path",
+    required=False,
+    default=None,
+    type=click.Path(dir_okay=True, file_okay=False, resolve_path=True, path_type=Path),
+    help="Directory path where to store the results of the static analysis.",
```
neilcampbell, line 144: Include where the default path is, assuming we always write the output to disk.

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+    *,
+    input_paths: tuple[Path],
+    recursive: bool,
+    force: bool,
+    baseline: bool,
+    output_path: Path | None,
+    detectors_to_exclude: list[str],
+) -> None:
+    """
+    Analyze the TEAL programs for common vulnerabilities.
```
neilcampbell, line 165:
```suggestion
    Analyze TEAL programs for common vulnerabilities using Tealer.
```

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+        if input_path.is_dir():
+            pattern = "**/*.teal" if recursive else "*.teal"
+            input_files.extend(sorted(input_path.glob(pattern)))
+        else:
+            input_files.append(input_path)
+    return sorted(set(input_files))
+
+
+@click.command(
+    name="analyze",
```
neilcampbell, line 104: If it's simple, it could be cool to support `analyse` as an alias.

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+                err=True,
+                fg="yellow",
+            )
+            continue
+
+        filename = generate_report_filename(file)
+
+        # If baseline is enabled, store the report in the snapshots folder otherwise store it in the reports folder
+        # If a custom output path is provided, store the report in the specified path
+        report_output_root = TEALER_SNAPSHOTS_ROOT if baseline else output_path or TEALER_REPORTS_ROOT
```
neilcampbell, line 196: Is `TEALER_SNAPSHOTS_ROOT` used for anything other than baseline reports? Potentially `TEALER_BASELINE_ROOT` is a more consistent name?

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,226 @@
...
+        if baseline and old_report:
+            has_diff = has_baseline_diff(
+                cur_file=cur_file, report_output_path=report_output_path, old_report=old_report
+            )
+            if has_diff:
+                raise click.exceptions.Exit(1)
+
+        reports[str(report_output_path.absolute())] = json.load(report_output_path.open())
+
+    table_rows = generate_table_rows(reports)
```
neilcampbell, line 220:
```suggestion
    summaries = generate_summaries(reports)
```

src/algokit/core/tasks/analyze.py
```diff
@@ -0,0 +1,187 @@
...
+        logger.error(f"Diff detected in {cur_file}! Please check the content of " f"{report_output_path}.")
+
+        return True
+
+    return False
+
+
+def generate_table_rows(reports: dict) -> dict[Path, list[list[str]]]:
+    """
+    Generate table rows from the reports dictionary.
```
neilcampbell, line 149: Tweak this wording to match any rename

src/algokit/cli/tasks/analyze.py
aorumbayev: Removed, since input is now using arguments not options, so if both file and recursive is set it prints a warning that recursive is ignored

src/algokit/cli/tasks/analyze.py
aorumbayev: Yes

src/algokit/cli/tasks/analyze.py
aorumbayev: Slightly tweaked the content

src/algokit/cli/tasks/analyze.py
aorumbayev: done

src/algokit/cli/tasks/analyze.py
aorumbayev: Can still rename it back, but now its the main path that i used for persisting the reports

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,230 @@
...
+    Check if the file contains template variables.
+
+    Args:
+        path (Path): The file path to check.
+
+    Returns:
+        bool: True if template variables are found, False otherwise.
+    """
+    content = path.read_text()
+    return bool(re.search(r"\bTMPL_\w*\b", content))
```
neilcampbell, line 58: Our beaker smart contract template generates TEAL with a comment like `// TMPL_DELETABLE` and `// TMPL_UPDATABLE`. Won't these be detected as incorrectly having template variables?

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,230 @@
...
+            click.secho(
+                f"Warning: Skipping {file} due to template variables. Substitute them before scanning.",
+                err=True,
+                fg="yellow",
+            )
+            continue
+
+        filename = generate_report_filename(file, duplicate_files)
+
+        # If diff_only is enabled, store the report in the snapshots folder otherwise store it in the reports folder
```
neilcampbell, line 194:
```suggestion
```
Not needed anymore

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,230 @@
...
+            if diff_only and old_report:
+                has_diff = has_baseline_diff(
+                    cur_file=cur_file, report_output_path=report_output_path, old_report=old_report
+                )
+                if has_diff:
+                    report_output_path.write_text(json.dumps(old_report.model_dump(by_alias=True), indent=2))
+                    raise click.exceptions.Exit(1)
+
+            reports[str(report_output_path.absolute())] = json.load(report_output_path.open())
+        except Exception as e:
```
neilcampbell, line 213: If an exception is thrown and diff_only is true, we probably want to revert the report back to old_report, as tealer may have written the current report to disc.

neilcampbell (approved):

src/algokit/cli/tasks/analyze.py
```diff
@@ -0,0 +1,230 @@
...
+        # If a custom output path is provided, store the report in the specified path
+        report_output_root = output_path or TEALER_SNAPSHOTS_ROOT
+        report_output_path = report_output_root / filename
+
+        command = generate_tealer_command(cur_file, report_output_path, detectors_to_exclude)
+        old_report = load_tealer_report(str(report_output_path)) if report_output_path.exists() and diff_only else None
+        try:
+            run_with_animation(run_tealer, f"Analyzing {index + 1} out of {total_files} files", command)
+
+            if diff_only and old_report:
```
neilcampbell, line 204: Currently if we have `diff_only = True` and `old_report=None`, the `has_baseline_diff` will not be run and as a result the process will exit with code 0. Essentially it will behave as if the `--diff` flag was not supplied.

I think we want to exit with code 1 if `diff_only = True` and `old_report=None` and not even run tealer in this instance.

PR #373, opened 2023-12-20
feat: Add support for a customisable named localnet

negar-abbasi:
## Proposed Changes

  - Add the --name option to the `Algokit Localnet start` to create a directory with name of sandbox_{name} that allows developers to start the localnet with their own configurations.
  - The rest of the LocalNet commands will work on the running LocalNet instance.
  - `algokit goal` command will also be executed in the running LocalNet instance.
  - There will be only one running LocalNet at a time.


src/algokit/core/sandbox.py
```diff
@@ -192,6 +195,18 @@ def check_docker_compose_for_new_image_versions(self) -> None:
                 "algod has a new version available, run `algokit localnet reset --update` to get the latest version"
             )
 
+    def save_current_directory(self) -> None:
+        current_dir_file = get_app_config_dir() / "localnet_directory.txt"
```
neilcampbell, line 199: Potentially we could use docker itself to track this state via `docker compose ls --all --filter name='algokit_sandbox*' --format json`. It would be fairly easy for this state file to get out of sync if the user interacts directly with docker.

src/algokit/cli/localnet.py
```diff
@@ -57,14 +58,27 @@ def localnet_group() -> None:
...
+    "name",
+    "--name",
+    default="sandbox",
+    help="a name for your localnet",
+)
+def start_localnet(name: str) -> None:
+    sandbox = ComposeSandbox(dir_name=f"sandbox_{name}" if name != "sandbox" else name)
+    sandbox.save_current_directory()
     compose_file_status = sandbox.compose_file_status()
     sandbox.check_docker_compose_for_new_image_versions()
```
neilcampbell, line 71: Do we want to run this when using a named localnet? The message printed will tell the user to use `algokit localnet reset --update`, which if they run that we'll prompt to confirm environment override.

Additionally this will always check against the `latest` tag, which in the case of a named localnet, the user may lock the image in the compose file to some other tag.

I think a better option is to notify the user on start of a named environment (alongside print the config path), that we won't be checking for updates in this localnet.

src/algokit/cli/goal.py
```diff
@@ -47,7 +47,7 @@ def goal_command(*, console: bool, goal_args: list[str]) -> None:
             "See https://docs.docker.com/get-docker/ for more information."
         ) from ex
 
-    sandbox = ComposeSandbox()
+    sandbox = ComposeSandbox(dir_name="sandbox")
```
neilcampbell, line 50: I think we may need to think through this a bit more, as it's possible the running localnet is not the default one.

neilcampbell (approved):

PR #382, opened 2024-01-09
ci: native binaries 

aorumbayev:
# Proposed changes

- Implement decision from ADR -> https://github.com/algorandfoundation/algokit-cli/pull/395
- Refines poetry installation using dedicated action and improved caching
- Removing redundant zenhub actions
- Separate workflow and action for building and appending binaries to release/prerelease
- Fixes several tests 
- Expanding build matrix for pytest workflow to test on windows/mac/ubunut against 3.10-3.12 python versions
- E2e tested on a complete fork of algokit-cli, see example of prerelease with artifacts of binaries appended https://github.com/aorumbayev/algokit-cli/releases/tag/v1.10.24-beta.3
- Initial portability test - to be expanded in more detailed set of tests once minor tweaks on cli are addressed

## Further notes

- ⚠️⚠️⚠️ this pr only adds support for building and dumping binaries into zipped archives on the release or prerelease, linking it with distribution over snap/winget/brew is a separate ADR that will follow up this PR
- Portability tests and remaining tweaks in cli to fix minor issues when cli is executed via binary is to be addressed in separate smaller scaled prs (issues are to be created)


joe-p:
One thing I think is worth considering, perhaps not initially but as a long-term goal, is releases on npm. I think it's safe to assume most Mac devs have brew installed, but the same is not true for windows developers and chocolatey. Having a way to install via brew, pipx, and npm covers a large portion of our developers. Below is a resource I found on distributing binaries through npm

https://blog.xendit.engineer/how-we-repurposed-npm-to-publish-and-distribute-our-go-binaries-for-internal-cli-23981b80911b

.github/actions/build-binaries/action.yaml
```diff
@@ -0,0 +1,67 @@
...
+  steps:
+    - name: Build Executable
+      run: ${{ inputs.build_command }}
+      shell: bash
+
+    - name: Test Executable
+      run: |
+        ls -l dist
+        ./dist/${{ inputs.package_name }}/${{ inputs.package_name }}${{ inputs.operating_system == 'windows-latest' && '.exe' || '' }} --help
+      shell: bash
```
robdmoore, line 31: Does this work on windows?

.github/workflows/build-binaries.yaml
```diff
@@ -0,0 +1,132 @@
...
+        uses: actions/checkout@v4
+        if: ${{ inputs.upload_binaries == 'true' }}
+        with:
+          fetch-depth: 0
+
+      - name: Checkout source code (for build)
+        uses: actions/checkout@v4
+        if: ${{ inputs.upload_binaries != 'true' }}
+        with:
+          fetch-depth: 1
```
robdmoore, line 24: Minor: Can we set fetch depth based on `inputs.upload_binaries` rather than duplicating the whole step?

.github/workflows/build-binaries.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.12"
```
robdmoore, line 29: Should we pass python version in?

.github/workflows/cd.yaml
```diff
@@ -38,17 +38,19 @@ jobs:
...
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
           python-version: "3.10"
-          cache: "poetry"
 
-      - name: Install dependencies
-        run: poetry install --no-interaction --no-root
+      - name: Set up Poetry
+        uses: abatilo/actions-poetry@v2
```
robdmoore, line 47: Are we happy trusting this?

robdmoore: What does it do differently than `pipx install poetry`?

.github/workflows/check-python.yaml
```diff
@@ -8,16 +8,21 @@ jobs:
...
     steps:
       - name: Checkout source code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
-      - name: Install poetry
-        run: pipx install poetry
-
-      - name: Set up Python 3.10
+      - name: Set up Python 3.12
```
robdmoore, line 13: This uses 3.12, but cd uses 3.10 - is that right?

.vscode/settings.json
```diff
@@ -2,7 +2,6 @@
   // General - see also /.editorconfig
   "editor.formatOnSave": true,
   "editor.codeActionsOnSave": {
-    "source.organizeImports": "explicit",
```
robdmoore, line 5: Why is this one removed?

robdmoore (approved):

.github/actions/build-binaries/action.yaml
aorumbayev: yes

.github/workflows/build-binaries.yaml
aorumbayev: wasn't able to templatize this bit - splitting them in two with conditional was intentional, looks like the value must be static

.github/workflows/check-python.yaml
aorumbayev: Yes, keeping cd as 3.10 to be on a safer side given that the cd wheel gets published to pypi

aorumbayev: Will revisit once follow up changes are merged

.vscode/settings.json
aorumbayev: Clashing with ruff sorting

aorumbayev:
@robdmoore i had to lock [chore: patching tealer](https://github.com/algorandfoundation/algokit-cli/pull/382/commits/49bd912c1336b6b678ef0f11b610efc1605e5565) tealer once again. Not really sure yet why the builds used to work prior to recent commits but started failing on 3.12 workers despite previous tests on 3.12. I suspect cash got invalidated and perhaps some related dependencies got refreshed so installing tealer starter breaking on 3.12 again. Tealer has a related repo - added a note to track it. For scope of this pr just gonna patch the ci cd - will open separate pr to rever installation of tealer from our fork on foundation org until tealer team patches the 3.12 completely (we contributed initial efforts as part of 0.1.1 already but looks like it was not enough). 

https://github.com/crytic/tealer/pull/209 

Otherwise all comments addressed and retested with release on a fork - merging now

Issue #383, opened 2024-01-10
Evaluate options for producing a native binary


neilcampbell:
## Description

As an AlgoKit developer
I want to evaluate the various options for [producing a native binary](https://github.com/algorandfoundation/algokit-cli/issues/384)
So that I'm informed of the tradeoffs of each approach

## Technical Details

Options to evaluate:
- PyInstaller
- cx_Freeze
- PyOxidizer
- Nuitka

Environments to Support:
- Windows (10+)
- MacOS (-1 from current)
- Linux - RPM Based - CentOS/RHEL
- Linux - Debian Based - Ubuntu/Fedora/Qubes OS (16.0.4+)
- Linux - Arch

Architectures to Support:
- x86
- x86_64
- arm64
- Apple silicon

## Acceptance Criteria

- A representative proof of concept has been built to assess suitability of each option
- Any findings/limitations/trade offs/problems that have been discovered are documented, allowing us to critically assess the options

## Tasks

- [x] Create a PyInstaller PoC @negar-abbasi
- [x] Create a cx_Freeze PoC
- [x] Create a PyOxidizer PoC
- [x] Create a Nuitka PoC

aorumbayev:
separate adr on distribution via snap, brew, winget is to follow

PR #395, opened 2024-01-16
docs: adr on native binaries

aorumbayev:
Addressing https://github.com/algorandfoundation/algokit-cli/issues/383

## Proposed Changes

  - evaluating switch to new distribution model with native binaries


aorumbayev:
@achidlow @daniel-makerx @robdmoore (cc @joe-p) please note this is still a draft WIP but feel free to leave some initial thoughts opinions on this

docs/architecture-decisions/2024-01-13_native_binaries.md
```diff
@@ -0,0 +1,139 @@
...
+    A -->|Wheel Build| K[Poetry]
+    K --> L[PyPi]
+    L -->|pipx| M[Python Users]
+```
+
+> Note that PyPi in the diagram above represents the existing distribution model for AlgoKit CLI prone to environment-specific bugs.
+
+The scope of this ADR only concerns the packaging for the CLI. The distribution via `snap`, `winget` and etc will be handled separately/in-parallel after decision and implementation of this ADR is in place.
+
+## Requirements
```
robdmoore, line 36: No need to install Python on your system if you aren't using a Python project.

docs/architecture-decisions/2024-01-13_native_binaries.md
```diff
@@ -0,0 +1,139 @@
+# AlgoKit CLI Enhancements
```
robdmoore, line 1:
```suggestion
# AlgoKit CLI native binaries
```

docs/architecture-decisions/2024-01-13_native_binaries.md
```diff
@@ -0,0 +1,182 @@
...
+- Active development and community support
+- Fairly fast build time via ci - ~3-4 minutes
+- Fairly small executable size (see benchmarking results below)
+- Marginally equal executable load time in `onedir` mode compared to `onefile` mode
+
+**Cons**
+
+- Occasionally requires manual configuration for more complex packages
+- Requires complex build packaging matrix to support multiple platforms and architectures
+- Requires minor tweaks in algokit cli to account for the fact that features relying on `sys.executable` will point to algokit cli executable instead of python interpreter. This is a minor change and can be done in a backwards compatible way however still a con to consider.
```
robdmoore, line 64: Is there a `sys.executable` equivalent that works?

docs/architecture-decisions/2024-01-13_native_binaries.md
```diff
@@ -0,0 +1,182 @@
...
+
+The benchmarking was performed on a MacBook M2 running macOS 14.2.1 and an ARM based Ubuntu 20.04.3 LTS running on a Parallels Desktop on the same machine.
+
+#### Results
+
+| Method              | macOS M2 | Ubuntu 20 ARM Linux VM | Windows 11 ARM |
+| ------------------- | -------- | ---------------------- | -------------- |
+| nuitka_onefile      | 3.634    | 1.465                  | 3.874          |
+| nuitka_onedir       | 0.2515   | 0.6200                 | 0.5136         |
+| pyinstaller_onedir  | 0.3228   | 0.7927                 | 0.6668         |
```
robdmoore, line 131: Given simplicity and lower build time, this performance is pretty on par with nuitka_onedir?

docs/architecture-decisions/2024-01-13_native_binaries.md
aorumbayev: @robdmoore we can still invoke anything within copier tasks so far i was thinking as a fix for this we can work on extending algokit cli to include something like `algokit run` which can act very similarly to `algokit deploy` and using that to wrap and invoke the inject.py properly. Plus we probably will cut dependency on inject.py execution via copier as part of work on init v2 and something like algokit run might get handy there as well.

docs/architecture-decisions/2024-01-13_native_binaries.md
```diff
@@ -0,0 +1,182 @@
...
+
+- Nuitka translates Python code into C and then compiles it, which can result in performance improvements.
+- Cross-Platform: Supports multiple platforms including Windows, macOS, and Linux.
+- More cross compilations options than PyInstaller
+- Official github action simplifies the process of building executables for different platforms.
+
+**Cons**
+
+- Compilation Time: The process of converting Python to C and then compiling can be time-consuming. Up to ~30 minutes on github with 3 parallel jobs.
+- Size of Executable: The resulting executables can be larger due to the inclusion of the Python interpreter and the compiled C code (see benchmarking results below).
```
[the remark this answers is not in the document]
aorumbayev: When you compile a Python script with Nuitka, it translates the Python code into C code that makes calls to the libpython functions. The C code is then compiled with a C compiler to produce an executable. This executable is able to run on a system that has the compatible version of libpython installed, or it can be compiled in standalone mode, which includes all necessary parts of the Python interpreter and any required modules within the binary itself, allowing it to run independently of a Python installation.

docs/architecture-decisions/2024-01-13_native_binaries.md
```diff
@@ -0,0 +1,182 @@
...
+- Does not support Python 3.12.
```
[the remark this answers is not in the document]
aorumbayev: @robdmoore so far we pre discussed this with @daniel-makerx and presumably bundling/packging puya would be done separately so that cli could lazy load it on initial request if user needs the compiler. So even though nuitka doesn't support it, if we go with pyinstaller - puya could reuse the setup from cli by the time this work will start implementation

docs/architecture-decisions/2024-01-13_native_binaries.md
aorumbayev: @robdmoore yes id say nuitka_onedir, pyinstaller_onedir and algokit via pypi is within a margin of milliseconds - still figured its worth outlining that nuitka indeed performs fastest given that its precompiled into C. From usage perspective though pyinstaller_onedir is comparably fast so wouldn't consider nuitka a huge advantage in terms of speed

docs/architecture-decisions/2024-01-13_native_binaries.md
```diff
@@ -0,0 +1,183 @@
...
+- Requires minor tweaks in algokit cli to account for the fact that features relying on `sys.executable` will point to algokit cli executable instead of python interpreter. This is a minor change and can be done in a backwards compatible way however still a con to consider.
+- Requires minor tweaks in algokit cli to introduce `multiprocessing.freeze_support()` to avoid issues with `vanity-address` task when executing via binary.
+
+#### PoC
+
+The PoC is available [here](https://github.com/algorandfoundation/algokit-cli/pull/393). It outlines a simple github action with extra setup that compiles algokit cli as a single file executable on latest versions of Windows, Mac and Linux github runners.
+
+### Option 3 - Cross platform python wheels with `cibuildwheel`
```
aorumbayev, line 93: Question for @achidlow @daniel-makerx , technically our pyproject.toml currently states that the build wheel is a pure python wheel. However we rely on dependencies that rely on non pure python based wheels like `cffi`, `cryptography`, `pycryptodomex` - however they all are distributed via manylinux build wheels and cover large amount of architectures/oses and etc. Does this imply that we can safely assume that algokit cli wheel is still pure python or reliance on packages as mentioned, implies we still need to build our own wheels in with tools like `cibuildwheel`? I can adjust this option to instead propose this tool https://github.com/cloud-custodian/poetry-plugin-freeze which we can use to simply freeze the wheel (does some tweaking with locking all specified dependencies prior to building the wheel) and we don't need to worry about manylinux stuff (assuming we are indeed building a pure python wheel)

aorumbayev: @achidlow i removed mentions of cibuildwheel as we clarified during the sync that algokit cli is indeed a pure python wheel

docs/architecture-decisions/2024-01-13_native_binaries.md
aorumbayev: @achidlow @daniel-makerx let me know what you think about this as well. 

To recap, the problem as highlighted in the adr is due to the fact that when running copier on full stack from a binary, the `sys.executable` leads to the binary not the python interpeter bundled with that binary. What do you think about introducing `algokit run` to be used to wrap execution of python files (or similarly to deploy, any commands). That way we can specify something like `algokit run 'python inject.py'` in the copier tasks instead of relying on sys.executable aliased variable named `_copier_python`

robdmoore (approved):

PR #396, opened 2024-01-23
fix: installation process for tealer (windows compatibility)

aorumbayev:
# Proposed changes

- Fixes compatibility with windows and aligns installation process for tealer to match the approach with poetry installation in bootstrap 
- since pytest is run via `pytest-xdist` in order to avoid race conditions on initial install of tealer, contributing docs and pipeline are updated to install tealer prior to execution of the pytest suite.

### DIFF
--- a/.github/workflows/build-python.yaml
+++ b/.github/workflows/build-python.yaml
@@ -25,7 +25,7 @@ jobs:
           cache: "poetry"
 
       - name: Install dependencies
-        run: poetry install --no-interaction
+        run: poetry install --no-interaction && pipx install tealer==0.1.1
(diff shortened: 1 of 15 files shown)


.github/workflows/build-python.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
           cache: "poetry"
 
       - name: Install dependencies
-        run: poetry install --no-interaction
+        run: poetry install --no-interaction && pipx install tealer==0.1.1
```
robdmoore, line 28: Do we need the version number hardcoded?

Issue #402, opened 2024-01-30
Improve AlgoKit CLI compatibility with execution via pyinstaller binaries


aorumbayev:
## Description

AlgoKit cli when packaged as a pyinstaller binary, has a set of issues when executed in that mode.

1. `algokit task vanity-address` requires patching to encorporate `multiprocessing.freeze_support()` to be enabled in `__main__.py`  to prevent execution failures when running in binary.
2. `algokit init` fails on any copier task execution. Requires aliasing of `algokit deploy` command into `algokit run` command that can be invoked from a template instead of relying on `_copier_python` variable. Cli should figure out the right user's default python interpreter to pass the invocation to via subprocess.

## Technical Details

- It is advisable to implement a helper method inside cli that serves to identify whether algokit is invoked via binary or via invocation of the wheel.

## Acceptance Criteria

- `algokit task vanity-address` command works in binary execution mode
- `algokit init` works on any copier template relying on _tasks where invocation of python script is invoked via `algokit run`
- Documentation for algokit run is available on the cli repository
- Updated CI/CD pipelines by introducing portability tests that verify main algokit cli flows by testing freshly build binary artifacts

## Tasks

- [ ] - patched vanity-address command in binary mode
- [ ] - patch copier tasks compatibility in binary mode
- [ ] - documented `algokit run`

---

TODO Refine conventions 

The hierarchy so far:
1. Milestone/Outcome - defined on an individual repo (case sensitive) 
2. Features - defined as github issues - the specific type identified by tag named `feature`
3. Task - defined as a github issue - serves as a smallest measurable unit of work - identified by tag named `task`

PR #406, opened 2024-02-05
fix: binary execution mode compatibility 

negar-abbasi:
- [#402](https://github.com/algorandfoundation/algokit-cli/issues/402)
- [#407](https://github.com/algorandfoundation/algokit-cli/issues/407)

### 402 
  - Implementing `multiprocessing.freeze_support()` to prevent execution failures when running `algokit task vanity-address` in binary.
 - Addressing the issue where algokit init fails during the execution of copier tasks in the fullstack template. The proposed solution involves using the system's Python path as a temporary substitute for `_copier_python`. This change is temporary until the issue with copier is resolved.

### 407
- Adding portability tests against `doctor`, `init`, `vanity-address` and `help` commands. 
- Minor bumps in versions of several github actions (setup-python to v5, download-artifacts to v4 and etc). Fixes several minor deprecation warnings in github workers logs
 
## Further notes

- Make sure to merge [fullstack](https://github.com/algorandfoundation/algokit-fullstack-template) PR first (marking it as a breaking change with a bump in min algokit version requirement)


src/algokit/cli/init.py
```diff
@@ -216,6 +217,15 @@ def init_command(  # noqa: PLR0913
     # provide the directory name as an answer to the template, if not explicitly overridden by user
     answers_dict.setdefault("project_name", directory_name)
 
+    system_python_path = next(get_python_paths(), None)
+    if system_python_path is not None:
```
aorumbayev, line 221: @achidlow @negar-abbasi - what if we expand this and also add usage of `is_binary_mode` and if its TRUE we execute a python script using something like [this](https://github.com/pyinstaller/pyinstaller/issues/2490) via `exec` i believe it should refer to currently used interpreter which in case of pyinstaller will be the one bundled along with the bootloader. That way users of binary installed algokit won't need to have python to run custom tasks in copier

achidlow: I think the solution in this PR is simpler. Adding another alogkit command which is called from a copier template `_tasks` to execute a Python script that is part of template's data is a very tight coupling compared to informing a template of a Python interpreter location on the `$PATH` via a supplied answer.

src/algokit/__main__.py
```diff
@@ -1,3 +1,10 @@
+from multiprocessing import freeze_support
+
 from algokit.cli import algokit
+from algokit.cli.common.utils import is_binary_mode
+
+# Required to support full feature parity when running in binary execution mode
+if is_binary_mode():
```
achidlow, line 7: It seems from the docs we can omit this `if` check and just always run `freeze_supprt`? https://docs.python.org/3/library/multiprocessing.html#multiprocessing.freeze_support

src/algokit/cli/init.py
aorumbayev: @achidlow i see so this is all handled by `get_python_paths`, if `which` did not return `python` or `python3` it will grab the `sys.executable` which will point to the interpreter bundled in binary (when running in binary mode).

aorumbayev:
Hey @negar-abbasi, i figured i'll combine this with portability tests issue (will update pr description later), still need a bit more time to refactor the pipelines a bit, i have wip progress on a fork will finalize this pr tomorrow for @neilcampbell and yourself to review. In the meantime feel free to proceed with other items

src/algokit/cli/init.py
```diff
@@ -216,6 +217,12 @@ def init_command(  # noqa: PLR0913
     # provide the directory name as an answer to the template, if not explicitly overridden by user
     answers_dict.setdefault("project_name", directory_name)
 
+    system_python_path = next(get_python_paths(), None)
```
neilcampbell, line 220: Should we do the same for the generators? It's possible a generator might try to execute a python script via copier.

tests/init/test_init.py
```diff
@@ -685,6 +707,57 @@ def test_init_with_custom_env(tmp_path_factory: TempPathFactory) -> None:
     )
 
 
+def test_init_fullstack_template_fails_on_missing_python(
```
neilcampbell, line 710: This isn't technically the fullstack template.

You could also add a comment to indicate that the fullstack template is an example of a template which executes python as part of running copier.

```suggestion
def test_init_template_with_python_fails_on_missing_python(
```

tests/init/test_init.py
```diff
@@ -685,6 +707,57 @@ def test_init_with_custom_env(tmp_path_factory: TempPathFactory) -> None:
...
+        ],
+        cwd=dummy_template["cwd"],
+        input="y\n",
+    )
+
+    assert result.exit_code == 1
+    verify(result.output, scrubber=make_output_scrubber())
+
+
+def test_init_fullstack_template_works(which_mock: WhichMock, dummy_template: dict[str, Path]) -> None:
```
neilcampbell, line 739:
```suggestion
def test_init_template_with_python_task_works(which_mock: WhichMock, dummy_template: dict[str, Path]) -> None:
```

.github/actions/build-binaries/action.yaml
```diff
@@ -40,22 +34,31 @@ runs:
...
+    - name: Upload binary as artifact (release)
       if: ${{ inputs.upload_binaries == 'true' }}
       uses: actions/upload-artifact@v4
       with:
-        name: algokit-cli-${{ inputs.operating_system }}-py${{ inputs.python_version }}
+        name: algokit-${{ inputs.operating_system }}-py${{ inputs.python_version }}
         path: algokit-${{ env.RELEASE_VERSION }}-${{ inputs.operating_system }}-py${{ inputs.python_version }}.tar.gz
 
+    - name: Upload binary as artifact (dev)
+      if: ${{ inputs.upload_binaries != 'true' }}
```
neilcampbell, line 56: @aorumbayev Should this input be renamed to something else?
Also should we be flagging these artefacts as pre-release?

aorumbayev (approved):
@neilcampbell approving Negar's bits but lmk if you have any further comments for the portability tests changes

.github/actions/build-binaries/action.yaml
```diff
@@ -24,41 +24,61 @@ runs:
       run: ${{ inputs.build_command }}
       shell: bash
 
-    - name: Test Executable
+    - name: Set TEST_CLI_PATH
       run: |
-        ls -l dist
-        ./dist/${{ inputs.package_name }}/${{ inputs.package_name }}${{ inputs.operating_system == 'windows-latest' && '.exe' || '' }} --help
+        ls -la dist/algokit/
```
neilcampbell, line 29: Any reason to keep this?

src/algokit/cli/init.py
neilcampbell: @negar-abbasi Unmarked this as resolved, as it doesn't appear like this has been handled.

negar-abbasi: I thought we are handling it with docs, telling user if they need this option then they have to do use 'python_path' convention as we did.

neilcampbell: But we won't be resolving the `python_path` inside the `generate` command, so it won't work inside a generator.

.github/actions/build-binaries/action.yaml
aorumbayev: @neilcampbell was useful for debugging but not anymore, can always revert back if needed

src/algokit/cli/init.py
aorumbayev: @negar-abbasi @neilcampbell see my last commit that adds the same logic to generate (i am not covering other default answers yet only the python_path answer for now). And there is an extra snapshot test that runs generate without mocking and tests that there is non zero exit code.

tests/init/test_init.py
```diff
@@ -714,10 +714,7 @@ def test_init_template_with_python_task_fails_on_missing_python(
     verify(result.output, scrubber=make_output_scrubber())
 
 
-def test_init_template_with_python_task_works(
-    which_mock: WhichMock, dummy_algokit_template_with_python_task: dict[str, Path]
-) -> None:
-    which_mock.remove("python")
```
aorumbayev: @negar-abbasi i haven't seen any effects of removing this i assume this was just a typo?

neilcampbell (approved):

tests/init/test_init.py
aorumbayev: @negar-abbasi i saw you restored the change back - can you explain why we need which_mock.remove('python') bit in this test?

src/algokit/cli/init.py
achidlow: ah, we need to update that to return `None` if `sys.executable and sys.executable.endswith("alogkit")` or something like that, it's not a valid interpreter so it shouldn't be returned. This goes for templates here as well as where it's used for `pipx`

PR #409, opened 2024-02-07
chore: update multidict

### LINKED ISSUE #378: Python 3.12 support for a transitive dependency `multidict`
## Description

Certain windows users reported errors during installation via pipx where a transitive dependency fails during build and requires user to have `C++  Build Tools` installed. 

The root cause is the `multidic` dependency that `aiohttp` relies on, cli inherits those by relying on `auth0-python` package.

Dedicated issue on multidict can be tracked here https://github.com/aio-libs/multidict/issues/887.

This issue is to be resolved as soon as a proper support is introduced in the multidict dependency.

neilcampbell:
update multidict, which prevents the need for c++ build tools on windows

Fixes #378


aorumbayev (approved):
@neilcampbell lets merge this after the pr with binary portability tests to trigger the pipelines for binaries and test the behaviour with multidict version bump

neilcampbell:
@aorumbayev Yeah good idea

PR #410, opened 2024-02-12
chore: update to the latest client-generator-ts

neilcampbell:


src/algokit/core/typed_client_generation.py
```diff
@@ -13,7 +13,7 @@
 
 logger = logging.getLogger(__name__)
 
-TYPESCRIPT_NPX_PACKAGE = "@algorandfoundation/algokit-client-generator@^2.5.0"
+TYPESCRIPT_NPX_PACKAGE = "@algorandfoundation/algokit-client-generator@^3.0.0"
```
robdmoore, line 16: How worried should we be about the breaking changes?

I wonder if we should allow people to override the version of this as a cli argument just in case they need to pin an old version (or pre-emptively use a new version).

PatrickDinh: I think it's worth to review this breaking change. It was introduced in [this commit](https://github.com/algorandfoundation/algokit-client-generator-ts/commit/d3cca63d7d4340adb8b52d76ce6a7a79f84cbf8f). It changed state properties and method args to camel case.

Supporting an CLI argument to specify a version is a great idea.

neilcampbell: We could pass the `--preserve-names` option by default which would prevent any breaking changes to the generated client.

neilcampbell: If we were to take a version number as a cli argument, we'd need to ignore it if you're generating a python typed client, as that is referenced as a dependency. Doesn't quite feel right.

It's probably better in that case for someone to lock the dependency in their project directly and run via `@algorandfoundation/algokit-client-generator generate` or `algokitgen generate`, rather than using `algokit`.

PatrickDinh: >We could pass the --preserve-names option by default which would prevent any breaking changes to the generated client. We could then make the breaking change as part of releasing 2.0 along with Puya.

I'm quite new to the code but looking at it, I think `--preserve-names` flag selects the `preservingSanitiser`, and it is a behave differently to the old (2.6.x) `defaultSanitiser`.

IMHO we either introduce the breaking change here or wait till Puya 2.0.

neilcampbell: Yeah you're correct, with and without `--preserve-names` is a breaking change.

neilcampbell:
This is being reworked in https://github.com/algorandfoundation/algokit-cli/pull/455, so closing this one.

PR #416, opened 2024-02-15
chore: build binary for Apple Silicon

### LINKED ISSUE #389: MacOS users can easily install the native AlgoKit cli
## Description

As a MacOS user
I want to install the AlgoKit cli using brew
So that I have a seamless install and upgrade experience

## Acceptance Criteria

- The AlgoKit cli is available to install using brew
- The AlgoKit cli detects when an upgrade is available and the user can initiate an upgrade via brew
- Self update mechanism is implemented as per https://github.com/orgs/algorandfoundation/projects/1?pane=issue&itemId=44360946

## Tasks

Tasks
 - [x] Get familiar with codesign commands for mac os binaries. Enhance release ci pipeline to codesign the binary prior to release to brew tap (evaluate whether there is any value to instead parametrize the build artifact action to take care of codesigning step vs having dedicated step as part of the cd workflow). 
 - [x] Add support for apple silicon runners 
 - [x] Duplicate existing binary build action for mac to be run on a matrix that includes both arm64 and amd64 instead of current setup that relies on amd64 based macos-latest runner. This is to ensure we support apple silicon. Github has free runners on m1 available as of last week, refer to github actions docs for more info.
 - [x] Upgrade existing brew release flow by copying poc implementation from [fork commit](https://github.com/aorumbayev/algokit-cli/commit/2d25ea98e2eed6b0f8231ff1f2407706f2484c26) (sample [action run](https://github.com/aorumbayev/algokit-cli/actions/runs/7647995869) that publishes cask with binary in it instead of the wheel). Further polish and refine poc implementation by updating the cask definition to conditionally pick arm vs amd64 binary based on os architecture of the user during cask installation (see [refs](https://stackoverflow.com/questions/77629028/does-homebrew-cask-support-architecture-based-distribution-for-apple-silicon-and) on how to achieve that). 
 - [x] Explicitly verify brew behaviour when updating algokit installed from brew as a wheel vs a new brew version that is installed as a binary.  
 - [ ] Perform initial e2e release by incorporating signing using Apple developer certificates provided by @Loedn
 - [ ] Document new installation option on the main readme for mac and brew users.

PatrickDinh:
Fixes #389 

## Proposed Changes

  - Include the process architecture into the output binary.
  - Install `poetry` directly with `pip`.
  - Build an extra binary for Apple Silicon.

### DIFF
--- a/.github/actions/setup-poetry/action.yaml
+++ b/.github/actions/setup-poetry/action.yaml
@@ -1,20 +1,21 @@
 name: "Python Poetry Action"
 description: "An action to setup Poetry"
-inputs:
-  poetry-version:
-    description: "The version of poetry to install"
-    required: false
-    default: "latest"
 runs:
   using: "composite"
   steps:
-    - if: ${{ inputs.poetry-version == 'latest' }}
+    # A workaround for pipx isn't installed on M1 runner.
+    # We should remove it after this issue is resolved.
+    # https://github.com/actions/runner-images/issues/9256
+    - if: ${{ runner.os == 'macOS' && runner.arch == 'ARM64' }}
       run: |
         pip install poetry
       shell: bash
-    - if: ${{ inputs.poetry-version != 'latest' }}
+
+    - if: ${{ runner.os != 'macOS' || runner.arch != 'ARM64' }}
       run: |
-        pip install poetry==${{ inputs.poetry-version }}
+        pip install --user pipx
+        pipx ensurepath
+        pipx install poetry
       shell: bash
 
     - name: Get full Python version
@@ -26,4 +27,4 @@ runs:
       uses: actions/cache@v4
       with:
         path: ./.venv
-        key: venv-${{ hashFiles('poetry.lock') }}-${{ runner.os }}-${{ steps.full-python-version.outputs.full_version }}
+        key: venv-${{ hashFiles('poetry.lock') }}-${{ runner.os }}-${{ runner.arch }}-${{ steps.full-python-version.outputs.full_version }}
--- a/.github/actions/setup-poetry/action.yaml
+++ b/.github/actions/setup-poetry/action.yaml
@@ -8,17 +8,13 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - run: |
-        pip install --user pipx
-        pipx ensurepath
-      shell: bash
     - if: ${{ inputs.poetry-version == 'latest' }}
       run: |
-        pipx install poetry
+        pip install poetry
       shell: bash
     - if: ${{ inputs.poetry-version != 'latest' }}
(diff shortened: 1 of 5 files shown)


.github/actions/setup-poetry/action.yaml
```diff
@@ -8,17 +8,13 @@ inputs:
...
   using: "composite"
   steps:
-    - run: |
-        pip install --user pipx
-        pipx ensurepath
-      shell: bash
     - if: ${{ inputs.poetry-version == 'latest' }}
       run: |
-        pipx install poetry
+        pip install poetry
```
robdmoore, line 13: Why pip rather than pipx?

robdmoore (approved):

.github/workflows/build-binaries.yaml
```diff
@@ -15,14 +15,16 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest, macos-14]
```
robdmoore, line 18: Is there a more explicit way of requesting apple silicon? This isn't obvious :P If you can't we should add a comment to explain...

.github/actions/setup-poetry/action.yaml
PatrickDinh: I found that on Apple Silicon, even thought pipx was installed in the previous step, pipx command isn't found here
[The failed build](https://github.com/PatrickDinh/algokit-cli/actions/runs/7884052914/job/21512425531)
I see that pipx is only used here to install poetry so I skip pipx.

.github/workflows/build-binaries.yaml
PatrickDinh: Good idea, I'll update it.

.github/actions/setup-poetry/action.yaml
aorumbayev, line 13: @PatrickDinh is this parametrized so it only uses pip when its apple silicon? I think we should still stick to pipx for other pipelines - arm silicon will be adding pipx at some point so its rather an exceptional case

.github/workflows/build-binaries.yaml
```diff
@@ -37,6 +39,12 @@ jobs:
       - name: Set up Poetry
         uses: ./.github/actions/setup-poetry
 
+      - name: Setup poetry cache
```
aorumbayev, line 43: Any specific reason for putting this back cause i moved it into the setup poetry action in the previous pr

aorumbayev (changes requested):

aorumbayev (approved):

PR #431, opened 2024-02-26
Refactor build binaries workflow

### LINKED ISSUE #389: MacOS users can easily install the native AlgoKit cli

PatrickDinh:
## Proposed Changes
- Move the content of the build binaries action into build binaries workflow. Doing this so that we can easily add more parameters to platform specified build binaries actions.
- Rename some steps in the CD workflow

Fixes #389


.github/workflows/build-binaries.yaml
```diff
@@ -34,9 +35,85 @@ jobs:
       - name: Install dependencies
         run: poetry install --no-interaction
 
-      - name: Build & test binary
-        uses: ./.github/actions/build-binaries
+      - name: Set release version
```
PatrickDinh, line 38: The rest of this file was copied from `build-binaries/action.yaml`.

neilcampbell (approved):

PR #440, opened 2024-02-27
docs: minor patch in the badge

aorumbayev:
# Proposed changes

- Minor tweak in readme, patching the visitors badge
- Minor tweak in CD pipeline, ignore changes in .vscode, .idea, .md files and docs folder to avoid spamming with algokit releases when there are no user facing changes in-place to release

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -4,6 +4,11 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - "docs/**"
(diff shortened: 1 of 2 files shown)


aorumbayev:
@neilcampbell ended up using the tweak in pipeline itself using ignore paths for now , worth tweaking the approach with semantic releases separately as we will need to do some changes there anyway if we also want to use the latest v9.x version of that package

neilcampbell (approved):

PR #441, opened 2024-02-28
feat: add compile py command

### LINKED ISSUE #400: Puya integration into algokit-cli, removing the need to install them explicitly
## Description

The goal is to ensure that `algokit-cli` handles installation of `puya-py` automatically and lazy-loads them as per usage.

## Technical Details

Using `algokit compile` uses approach similar to `algokit generate` where semantics are inferred based on file extension. `.py` -> puya. An alternate approach is to use `algokit compile py`, so the compiler is known when the command is executed. This also has the added benefit of be extensible for future support of other compilers that may require different arguments.

We want to integrate this into the cli once 1.0 of Puya has released and update the cli to v2.

## Acceptance Criteria

- [ ] A user is able to execute the puya compiler via `algokit-cli` and turn their PuyaPy smart contract into a deployable teal contract, without explicitly installing it by executing `algokit compile *.py`
- [ ] A user is able to discover any compilation issues in their PuyaPy contract using `algokit compile`

## Tasks

- [ ] PuyaPy compiler is integrated into the cli, usage is tested & documented

PatrickDinh:
## Proposed Changes
- Add "compile py" command which compiles Python to TEAL with Puyapy
- Puya compilation errors are shown
image

Fixes #400


src/algokit/cli/compile.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@click.group("compile")
+def compile_group() -> None:
+    """Compile smart contracts to TEAL"""
+
+
+@compile_group.command(
+    "py",
+    short_help="Compile Python to TEAL with Puyapy",
```
PatrickDinh, line 18: Should point to puyapy document for the arguments

src/algokit/core/compile.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def _find_puya_command_at_version(version: str) -> list[str]:
+    """
+    Find puya command with a specific version.
+    If the puya version isn't installed, install it with pipx run.
+    """
+    for puyapy_command in _get_candidates_puyapy_commands():
+        try:
+            puyapy_version_result = run([*puyapy_command, "--version"])
```
PatrickDinh, line 18: Puya doesn't support `--version` yet, we catch it here
```
if puyapy_version_result.exit_code == 0
```
That means, if the user requests for a specific puya version, `pipx run` is always used, until puya support `--version` option.

aorumbayev (commented):
Great work @PatrickDinh , left some comments - also to clarify docs are removed temporarily? The plan is to rewrite it again once this goes public? Wonder if we should have the docs written beforehand and stashed somewhere internally to not spend time rewriting it once we make it public.

PatrickDinh:
> Great work @PatrickDinh , left some comments - also to clarify docs are removed temporarily? The plan is to rewrite it again once this goes public? Wonder if we should have the docs written beforehand and stashed somewhere internally to not spend time rewriting it once we make it public.

Thanks @aorumbayev for reviewing, good call out about the doc, I'll stash it locally.

src/algokit/core/compile.py
neilcampbell: FYI this has been added, it just hasn't been released to PyPi yet https://github.com/algorandfoundation/puya/pull/110

src/algokit/core/compile.py
```diff
@@ -0,0 +1,80 @@
...
+        except OSError:
+            pass  # in case of path/permission issues, go to next candidate
+        else:
+            if puyapy_version_result.exit_code == 0 and (
+                extract_version_triple(version) == extract_version_triple(puyapy_version_result.output)
+            ):
+                return puyapy_command
+
+    pipx_command = find_valid_pipx_command(
+        "Unable to find pipx install so that `PuyaPy` compiler can be installed; "
```
neilcampbell, line 28:
```suggestion
        "Unable to find pipx install so that the `PuyaPy` compiler can be installed; "
```

src/algokit/core/compile.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            puyapy_help_result = run([*puyapy_command, "-h"])
+        except OSError:
+            pass  # in case of path/permission issues, go to next candidate
+        else:
+            if puyapy_help_result.exit_code == 0:
+                return puyapy_command
+
+    pipx_command = find_valid_pipx_command(
+        "Unable to find pipx install so that `PuyaPy` compiler can be installed; "
```
neilcampbell, line 55:
```suggestion
        "Unable to find pipx install so that the `PuyaPy` compiler can be installed; "
```

There is a bit of duplicate here in the failure message, that we can extract into a constant.

src/algokit/core/compile.py
```diff
@@ -0,0 +1,80 @@
+from collections.abc import Iterator
```
neilcampbell, line 1: I'm wondering if we should setup the structure to support additional compilers now?

Similar to what we've done with `task`.
I guess we'd have a directory `compile` inside both `cli` and `core`, which has a `py.py` file? Worth also double confirming that we are happy with `py` and don't want to use `python`?

src/algokit/core/compile.py
```diff
@@ -0,0 +1,80 @@
...
+                return puyapy_command
+
+    pipx_command = find_valid_pipx_command(
+        "Unable to find pipx install so that `PuyaPy` compiler can be installed; "
+        "please install pipx via https://pypa.github.io/pipx/ "
+        "and then try `algokit compile py ...` again."
+    )
+
+    return [
+        *pipx_command,
```
neilcampbell, line 34: Does pipx check python version constraints when running or installing a dependency?
It would be good to confirm we gracefully handle when the user is running Python < 3.12.

Poetry I think is a different story, as it should use the python in their venv, which I'm assuming should be 3.12 if they've installed Puya.

PatrickDinh: `pipx install puya` throws an exception if the python version on the user's machine is < 3.12. It isn't easy to handle it nicely because we don't know which python version is used by pipx. I suggest that we throw the pipx exception (the current behaviour).

neilcampbell (approved):
Looks good.
Might be worth just confirming internally if we want `algokit compile py` or `algokit compile python`?

src/algokit/cli/compile.py
```diff
@@ -0,0 +1,61 @@
...
+    run_result = run(
+        [
+            *puya_command,
+            *puya_args,
+        ],
+    )
+    click.echo(run_result.output)
+
+    if run_result.exit_code != 0:
+        raise click.exceptions.Exit(run_result.exit_code)
```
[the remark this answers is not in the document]
aorumbayev: @PatrickDinh so i assume this is red by default, handled by click and the fact its an exception?

tests/compile/test_py.py
```diff
@@ -0,0 +1,93 @@
...
+
+    mocker.patch("algokit.core.proc.Popen").side_effect = proc_mock.popen
+
+    result = invoke(f"compile py {_normalize_path(hello_world_contract_path)}")
+
+    assert result.exit_code == 0
+    verify(result.output)
+
+
+def test_puyapy_is_installed_globally(hello_world_contract_path: Path, mocker: MockerFixture) -> None:
```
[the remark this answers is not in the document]
aorumbayev: @PatrickDinh sorry if im missing something but looks like all tests are still mocking calls to the puypy - have you decided to opt in for mocking everything puya specific?

aorumbayev (commented):
@PatrickDinh just minor comments left regarding my initial suggestions with tests, lmk what you think

src/algokit/cli/compile.py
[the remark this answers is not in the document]
PatrickDinh: In my test locally, without setting `fg="red"` click renders the error without color. Reading the [doc](https://click.palletsprojects.com/en/8.1.x/api/#click.secho), I think the flag `err=True` only writes the output to `stderr`.

tests/compile/test_py.py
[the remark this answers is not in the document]
PatrickDinh: I think renaming the file from `test_py` to `test_python` messed with GitHub and I couldn't see the new tests as well. I added them [here](https://github.com/algorandfoundation/algokit-cli/pull/441/files#diff-553a9ec78fe3643d061564d6c2a36e95708b08118df093cf1fe199731ef5f38bR109-R136). Let me know what you think.

src/algokit/cli/compile.py
```diff
@@ -0,0 +1,93 @@
...
+
+import click
+
+from algokit.core.compile.python import find_valid_puyapy_command
+from algokit.core.proc import run
+
+logger = logging.getLogger(__name__)
+
+
+class CompileGroup(click.Group):
```
aorumbayev, line 13: There is a lot here this isn't fully utilized, other samples of overwriting the group class usually inject dynamic commands or take care of aliasing. In this case if we want to have both py/python to be accessible there are a few simpler options:

1.  Using a simplified custom group class with only get_command overload
```
class AliasedGroup(click.Group):
    def get_command(self, ctx: click.Context, cmd_name: str) -> click.Command | None:
        rv = click.Group.get_command(self, ctx, cmd_name)
        if rv is not None:
            return rv

        if cmd_name == "py":
            return click.Group.get_command(self, ctx, "python")
          
         ... and so on for any ts -> typescript and other shorthands. 

        return None
```

In this case docstring in the python command need to include ref that its also callable from py:
```
Compile Python contract(s) to TEAL with PuyaPy. Additionally available via `algokit compile py` shorthand alias.
```
       
2. Adding duplicated python command to the compile group manually:
    
  ```
@click.command(
    "python",
    short_help="Compile Python to TEAL with PuyaPy",
    help="Compile Python to TEAL with PuyaPy, review https://github.com/algorandfoundation/puya for usage",
    context_settings={
        "ignore_unknown_options": True,
    },
    add_help_option=False,
)
@click.argument("puya_args", nargs=-1, type=click.UNPROCESSED)
@click.pass_context
def compile_py_command(context: click.Context, puya_args: list[str]) -> None:
    """
    Compile Python contract(s) to TEAL with PuyaPy
    """
    version = str(context.obj["version"]) if context.obj["version"] else None

    puya_command = find_valid_puyapy_command(version)

    run_result = run(
        [
            *puya_command,
            *puya_args,
        ],
    )
    click.echo(run_result.output)

    if run_result.exit_code != 0:
        click.secho(
            "An error occurred during compile. Ensure supplied files are valid PuyaPy code before retrying.",
            err=True,
            fg="red",
        )
        raise click.exceptions.Exit(run_result.exit_code)


compile_group.add_command(compile_py_command, "python")
compile_group.add_command(compile_py_command, "py")
  ```
  
  Last 2 lines achieve a simpler equivalent of what you currently have in CompileGroup class.

aorumbayev (approved):

Issue #271, opened 2023-05-29
Algokit with Podman


ksinkar:
I have installed podman on my system and set it up such that it can behave as a [drop-in replacement](https://thesynack.com/posts/docker-compose-podman/) for docker.

I am able to do
`sudo docker-compose up` and have systems up and running according to their *compose.yml* files.

However running
``` bash
algokit localnet start 
```
gives me an error saying that  Docker Compose not found. 

Is anyone successfully using algokit with podman?

robdmoore:
Closing due to inactivity, feel free to reopen if there is still a problem.

ShoGinn:
I know this was closed, but because of the tightly coupled nature of the docker (brand) integration, PodMan is a great alternative for those of us who are security-minded and want a non-root solution.

I'll take a look at the code to see if there is some workaround but I'm hopeful a contributor has the knowledge on how we can make this container agnostic.

robdmoore:
If you’re happy to advocate and assist we’re definitely happy to help

ShoGinn:
> If you’re happy to advocate and assist we’re definitely happy to help

Can do; let me dig into the code and ensure I'm in line with `algorand/sandbox` as well as other coding practices in `localnet.py` and `sandbox.py`

I think this should be an easy transition I just want to make sure I'm not missing anything before making the pull request.

PR #456, opened 2024-03-19
feat: GitHub Codespaces support in LocalNet command group

aorumbayev:
Spike following the https://github.com/orgs/algorandfoundation/projects/1/views/1?filterQuery=aorum&pane=issue&itemId=56476748

## Proposed Changes

This document serves as an informal Architecture Decision Record (ADR). Proceedings from the PR discussion will be converted to an ADR markdown file later.

### Motivation

A significant number of users attending bootcamps encountered scenarios where their devices had hardware incompatibilities with Docker (primarily certain Windows 10 variations). The proposal aims to provide a convenient built-in experience that leverages GitHub CLI capabilities to provision a dummy Codespace instance (using the Algokit base template as a target) to port forward localnet ports to the user's machine.

### Proposal

The idea is to provide a new command under `localnet` group to cover this scenario for users experiencing issues with docker. We can rely on GitHub CLI to leverage codespaces and their port-forwarding capabilities, allowing the Algokit CLI to lazily load the dependency if not present. The current proof of concept assumes winget for Windows and uses a 'webi' executable on Unix devices.

On the first login, the user is redirected to GitHub CLI's interactive authentication flow. Afterward, any existing Codespaces with the 'sandbox' prefix are automatically deleted. Finally, a fresh Codespace is created, and after a 60-second sleep (a holistic estimate for base machine quota, roughly half for container boot, and the remaining quarters for `pipx algokit install` and starting the localnet), the indexer, algod, and kmd ports from the Codespace are redirected to the user's machine. The user can then open a new tab and interact with the localnet as if it were running locally (similar to the Docker-based experience).

### Pros

- One of the simplest ways to solve the main requirement of providing alternative mechanisms for users to access localnet if Docker is not an option. It's important to maintain the Docker-based approach as the recommended option, given that Docker still offers advantages in offline local development scenarios and abstract away heaps of 'native' complexity. However, a simple, interactive experience within the `localnet` command group that provides a free quota-based temporary on-demand instance can offer a flexible alternative solution for any edge case where Docker is unavailable for local development.
- Embraces the notion of dev containers, which are actively propagated by official Algokit templates. Provides a marginally similar experience to using the Docker-based localnet.
- GitHub CLI is available on a massive array of architectures and operating systems, both as native binaries and via major package managers, allowing conditional handling of main variations by picking the appropriate distribution channel to obtain the artifact.

### Cons and Minor Notes

- Cannot work offline.
- Similar to the Docker-based approach, for simplicity, we assume that only one Codespace-based localnet instance can be up and running (otherwise, default ports will clash). An advanced developer can always tailor the ports as needed, so it's more of an inconvenience for beginners.

### Next steps

- [ ] - address comments, gather feedback on the proposal. Which parts of the initial poc can be further refined/improved?
- [ ] - if this proposal or a modified version based on final discussion in the pr are accepted, the pr implementation is to be refined, covered with snapshot tests and documented. 

How to play around with the poc ?
`pipx install git+https://github.com/algorandfoundation/algokit-cli@feat/codespaces-localnet --force` then `algokit localnet codespace start`


ShoGinn:
@aorumbayev This is an epic feature; I've been trying to implement an offline container-agnostic approach since, as you stated, using Docker can be problematic.

My approach is similar to yours in that we need a local network engine. Thank you for not destroying the local network design. I'm adding this to my workboard, but this might be a better stopgap.

ShoGinn:
@aorumbayev I could not get the pull to work without docker installed... I'm guessing its in the localnet group of the cli erroring out.

neilcampbell:
@ShoGinn Nice fine, you're spot on. The presence of docker is checked when the localnet command group is loaded. At the moment this PR is a quick spike to see what is possible, so a bit rough around the edges.

aorumbayev:
@ShoGinn this is still a preview mvp, but i patched the issue you had with docker, feel free to retry again

ShoGinn:
> @ShoGinn this is still a preview mvp, but i patched the issue you had with docker, feel free to retry again

I actually tried it when I saw you made some commits. It works great, and I'm excited to see how it progresses.

This solution is much cleaner than local docker/podman and enables a solid dev environment.

I think the CLI is going to be the biggest work. My recommendation is that rather than having dappflow or other commands check for the engine, have it check for the URL: port, but that's just a recommendation.

aorumbayev:
@neilcampbell @robdmoore @Loedn , ill still keep pr description as adr up until this pr is closed/merged/resolved. 
For now assume this comment is the description (since initially it was a draft mvp, latest commits polish up the experience and add docs and tests)

# Proposed changes

- algokit localnet codespace command - a single command that takes care of one full lifecycle from creation to termination -(user can opt out from auto cleanup with force flag) of a github codespace where upon localnet being booted inside the container - algokit cli will rely on gh cli to port forward the kmd algod and indexer to local network. See docs for detailed breakdown of behaviour and/or pipx install the branch to try it out yourself 
- minor refinements, improving the start_with_animation spinner to gracefully handle keyboard interrupts. 
- initial batch of snap shot tests (TODO: more to be added tomorrow)
- initial documentation 

(cc @ShoGinn , feel free to check it out comment too, first hand user feedback is always welcome! :-) )

ShoGinn:
@aorumbayev I did some tests with the latest code; the codespace will start without start (I'm not sure if that is intended). `algokit localnet codespace`  I expected the cli to provide some instruction without start.

Also, because the cli is always looking for docker, it won't work for `algokit localnet explore` etc

I would recommend that localnet include codespace or docker for commands, dig deeper for things like logs, etc., and make the explore and console less restrictive, just naturally failing or checking ports rather than files.

aorumbayev:
@ShoGinn hey thank you for checking it out!

RE: starting without start
This is indeed intended behaviour changes in this [commit](https://github.com/algorandfoundation/algokit-cli/pull/456/commits/51458050722899f20cd6c8d54421b81d2af6a48e#diff-a282d0691bc02d5269f540c1f5197bc0d67126e584d990472bb0afa25259e62f). The reasoning was due to the fact that the command is mainly designed to work with on-demand codespaces that it creates (unless a -n name to a specific instance is provided in such cases it will try to just reconnect to an existing instance). There were no use cases for stop, reset and reset --update and having them integrated under docker based localnet command groups can put a lot more coupling with gh cli and we still want to maintain a fairly light set of features that cli wraps for the user (because gh cli by itself is very flexible and user friendly out of the box). The initial draft you tried was still an early mvp, so i'd say you can expect a lot less breaking changes until this pr is finalized. 

RE: explore
Good point! I will make sure it can work with both codespace and docker based modes. (will let you know once commit is in place)

RE: 'I would recommend that localnet include codespace or docker for commands' 
I got the second half of this but regarding this specific bit, can you expand a little bit more? What are you referring to when you are mentioning localnet? A command group within cli or the group of containers representing the localnet ?

ShoGinn:
> @ShoGinn hey thank you for checking it out!
> 
> RE: starting without start This is indeed intended behaviour changes in this [commit](https://github.com/algorandfoundation/algokit-cli/pull/456/commits/51458050722899f20cd6c8d54421b81d2af6a48e#diff-a282d0691bc02d5269f540c1f5197bc0d67126e584d990472bb0afa25259e62f). The reasoning was due to the fact that the command is mainly designed to work with on-demand codespaces that it creates (unless a -n name to a specific instance is provided in such cases it will try to just reconnect to an existing instance). There were no use cases for stop, reset and reset --update and having them integrated under docker based localnet command groups can put a lot more coupling with gh cli and we still want to maintain a fairly light set of features that cli wraps for the user (because gh cli by itself is very flexible and user friendly out of the box). The initial draft you tried was still an early mvp, so i'd say you can expect a lot less breaking changes until this pr is finalized.
> 
Makes sense!
> RE: explore Good point! I will make sure it can work with both codespace and docker based modes. (will let you know once commit is in place)
> 
Can't wait!
> RE: 'I would recommend that localnet include codespace or docker for commands' I got the second half of this but regarding this specific bit, can you expand a little bit more? What are you referring to when you are mentioning localnet? A command group within cli or the group of containers representing the localnet ?

Sorry this was a fragmented thought! I just meant in the cli make commands specific to the docker portion a sub-command in the cli.

```
  logs       See the output of the Docker containers
  reset      Reset the AlgoKit LocalNet.
  start      Start the AlgoKit LocalNet.
  status     Check the status of the AlgoKit LocalNet.
  stop       Stop the AlgoKit LocalNet.
```

All of these commands should be under `algokit localnet docker`

that way when implemented you know what features are what.

```
algokit localnet
Usage: algokit localnet [OPTIONS] COMMAND [ARGS]...

Options:
  -h, --help  Show this message and exit.

Commands:
  codespace  Manage the AlgoKit LocalNet in GitHub Codespaces.
  docker     Manage the Algokit LocalNet in local Docker environment
  console    Run the Algorand goal CLI against the AlgoKit LocalNet via a Bash console so you can execute multiple
             goal commands and/or interact with a filesystem.
  explore    Explore the AlgoKit LocalNet using Dappflow
```

aorumbayev:
@ShoGinn i see, i do agree that it does make the command group semantics a bit cleaner (given that this pr is now introducing an alternative mechanism to accessing the localnet).

1. Your option -> one way this could be done without breaking changes is applying the same process we did to algokit bootstrap and deploy commands at the root command level in latest v2 release. They are still there yet they are hidden from help and upon invoking you see a deprecation warning. However, that would still require an eventual change in majority of algokit repos to switch to new syntax (and any algokit based project relying on it). This is a bit of a broader suggestion, i wonder what are your opinions on this @robdmoore @Loedn @neilcampbell ?)

2. Alternative, moving the codespace under task command group. While having codespaces under localnet makes sense (and codespaces approach being part of a solution to a formal outcome related to localnet), we also have a separate command group category called tasks which we use exclusively for extra utils OR integrations with external tools (see `algokit task analyze` for instance). As an alternative we can also move this codespace command under algokit task command group. That wouldn't require a breaking change in a large majority of existing algokit repos and projects used to invoking localnet directly without extra docker command group. Curious on opinions for this option as well.

neilcampbell:
> Your option -> one way this could be done without breaking changes is applying the same process we did to algokit bootstrap and deploy commands at the root command level in latest v2 release. They are still there yet they are hidden from help and upon invoking you see a deprecation warning. However, that would still require an eventual change in majority of algokit repos to switch to new syntax (and any algokit based project relying on it). This is a bit of a broader suggestion, i wonder what are your opinions on this @robdmoore @Loedn @neilcampbell ?)

It's definitely an option. I do quite like that we are somewhat decoupled from Docker with the command semantics, as I guess it's still possible LocalNet could be an entirely native setup.

> Alternative, moving the codespace under task command group. While having codespaces under localnet makes sense (and codespaces approach being part of a solution to a formal outcome related to localnet), we also have a separate command group category called tasks which we use exclusively for extra utils OR integrations with external tools (see algokit task analyze for instance). As an alternative we can also move this codespace command under algokit task command group. That wouldn't require a breaking change in a large majority of existing algokit repos and projects used to invoking localnet directly without extra docker command group. Curious on opinions for this option as well.

A potential issue with this approach is discoverability. It's probably not the first place someone will look to run a hosted LocalNet. What would the command be in this case `algokit task codespace localnet`? It does allow for the extensibility of adding other potential codespace features in the future.

I think there are pros and cons of the approaches. Let's chat to work them out and make a decision.

aorumbayev:
@ShoGinn sorry it took a while to get back to this since we were focused on other higher priority items. 

Feel free to track the progress on this in the newly created github issue here, https://github.com/algorandfoundation/algokit-cli/issues/504. You can assume the codespaces to be available on algokit cli as soon as that issue is closed.

ShoGinn:
@aorumbayev Thanks for picking up the slack on the container agnostic setup (#271)! I've been busy with other projects and have yet to do this!

docs/features/localnet.md
```diff
@@ -16,6 +18,10 @@ If you are on Windows then you will need WSL 2 installed first, for which you ca
 
 Alternatively, the Windows 10/11 Pro+ supported [Hyper-V backend](https://docs.docker.com/desktop/install/windows-install/) for Docker can be used instead of the WSL 2 backend.
 
+### Podman support
+
+If you prefer to use [Podman](https://podman.io/) as your container engine, make sure to install and configure `podman` first. Then you can set the default container engine to use by executing `algokit localnet config podman`. See [Container-based LocalNet](#container-based-localnet) for more details.
```
neilcampbell, line 23: As discussed, let's move this to `algokit config localnet`

Will need to be updated everywhere.

docs/features/localnet.md
```diff
@@ -6,6 +6,8 @@ AlgoKit LocalNet uses Docker images that are optimised for a great dev experienc
 
 The philosophy we take with AlgoKit LocalNet is that you should treat it as an ephemeral network. This means assume it could be reset at any time - don't store data on there that you can't recover / recreate. We have optimised the AlgoKit LocalNet experience to minimise situations where the network will get reset to improve the experience, but it can and will still happen in a number of situations.
 
+> For details on executing `algokit localnet` without `docker` refer to the [codespaces](#codespaces) section.
```
neilcampbell, line 9:
```suggestion
> For details on executing `algokit localnet` without `docker` or `podman` refer to the [codespaces](#codespaces) section.
```

docs/features/localnet.md
```diff
@@ -120,3 +136,26 @@ AlgoKit Utils provides methods to help you do this:
 - Python - [`ensure_funded`](https://algorandfoundation.github.io/algokit-utils-py/html/apidocs/algokit_utils/algokit_utils.html#algokit_utils.ensure_funded) and [`get_dispenser_account`](https://algorandfoundation.github.io/algokit-utils-py/html/apidocs/algokit_utils/algokit_utils.html#algokit_utils.get_dispenser_account)
 
 For more details about the `AlgoKit localnet` command, please refer to the [AlgoKit CLI reference documentation](../cli/index.md#localnet).
+
+## GitHub Codespaces-based LocalNet
+
+The AlgoKit LocalNet feature also supports running the LocalNet in a GitHub Codespace with port forwarding by utilizing the [GitHub CLI](https://github.com/cli/gh). This allows you to run the LocalNet without the need to use Docker. This is especially useful for scenarios where certain hardware or software limitations may prevent you from being able to run Docker.
+
+To run the LocalNet in a GitHub Codespace, you can use the `algokit localnet codespace` command.
+By default without `--force` flag it will prompt you to delete stale codespaces created earlier (if any). Upon termination it will also prompt the user to delete the codespace that was used prior to termination.
```
neilcampbell, line 145:
```suggestion
By default without `--force` flag it will prompt you to delete stale codespaces created earlier (if any). Upon termination it will also prompt to delete the codespace that was used prior to termination.
```

src/algokit/cli/doctor.py
```diff
@@ -42,23 +43,25 @@ def doctor_command(*, copy_to_clipboard: bool) -> None:
...
-                "Docker required to run `algokit localnet` command; install via https://docs.docker.com/get-docker/"
-            ],
+        container_engine: check_dependency(
+            [container_engine, "--version"],
+            missing_help=[f"`{container_engine}` required to run `algokit localnet` command; install via {docs_url}"],
         ),
-        "docker compose": check_dependency(
-            DOCKER_COMPOSE_VERSION_COMMAND,
-            minimum_version=DOCKER_COMPOSE_MINIMUM_VERSION,
+        f"{container_engine} compose": check_dependency(
```
neilcampbell, line 58: Does the podman compose equivalent follow the same versioning as docker?

src/algokit/cli/doctor.py
```diff
@@ -42,23 +43,25 @@ def doctor_command(*, copy_to_clipboard: bool) -> None:
...
+            COMPOSE_VERSION_COMMAND,
+            minimum_version=compose_minimum_version,
             minimum_version_help=[
-                f"Docker Compose {DOCKER_COMPOSE_MINIMUM_VERSION} required to run `algokit localnet` command;",
-                "install via https://docs.docker.com/compose/install/",
+                f"{container_engine.capitalize()} Compose {compose_minimum_version} required to",
+                "run `algokit localnet` command;",
```
neilcampbell, line 63: This results in the output being split over 3 lines, rather than 2 (as before).

src/algokit/cli/doctor.py
aorumbayev: @neilcampbell no but the get_min_compose_version is dynamically resolving the value

src/algokit/cli/codespace.py
```diff
@@ -0,0 +1,113 @@
...
+@click.option(
+    "-i", "--indexer-port", default=8980, required=False, help="The port for the Algorand indexer. Defaults to 8980."
+)
+@click.option("-k", "--kmd-port", default=4002, required=False, help="The port for the Algorand kmd. Defaults to 4002.")
+@click.option(
+    "-n",
+    "--codespace-name",
+    default="",
+    required=False,
+    help="The name of the codespace. Defaults to random name with timestamp.",
```
neilcampbell, line 46:
```suggestion
    help="The name of the codespace. Defaults to ${CODESPACE_NAME_PREFIX}_timestamp.",
```

aorumbayev:
@neilcampbell switched to webi installer on windows too and the official installer recommends restarting the terminal session so ive matched the cli behaviour with that, if gh is not installed it will exit upon auto installation and user will need to re run the codespace command upon restart of the terminal (will update the docs on this tomorrow). 
Other than that added extra idle timeout parameter, just port forwarding without doing anything extra is considered idle as well so user can set up up to 240 minutes of active state before codespace is shut down, each time there is any sort of interaction with the forwarded localnet the timer will be reset.

tests/localnet/test_localnet_codespace.py
```diff
@@ -0,0 +1,133 @@
...
+    mocker: MockerFixture, proc_mock: ProcMock, tmp_path_factory: pytest.TempPathFactory
+) -> None:
+    cwd = tmp_path_factory.mktemp("cwd")
+    dummy_script_path = cwd / "webi_dummy_installer.ps1"
+    dummy_script_path.touch()
+
+    proc_mock.should_fail_on(
+        ["gh", "--version"],
+    )
+    proc_mock.set_output(["powershell", "--version"], ["PowerShell 7.2.1"])
```
neilcampbell, line 41: I don't believe that `powershell --version` is available.

neilcampbell: I also get the following error when the webi installer runs. It still installs though.

```
Set-ExecutionPolicy : The 'Set-ExecutionPolicy' command was found in the module 'Microsoft.PowerShell.Security', but the module could not be loaded. For more information, run 'Import-Module
```

aorumbayev: @neilcampbell should be resolved now, powershell is checked via different command that is more versatile. 

The Set-ExecutionPolicy on the other hand is a bit tricky to replicate, whats the output of `Get-ExecutionPolicy -List` on your side when you run it in powershell on windows (preferably same environment as the one where the error occurred during install).

neilcampbell: @aorumbayev Output is this:
Scope ExecutionPolicy
----- ---------------
MachinePolicy       Undefined
   UserPolicy       Undefined
      Process       Undefined
  CurrentUser       Undefined
 LocalMachine    RemoteSigned

neilcampbell: @aorumbayev Forgot to mention I was running `algokit -v localnet codespace`.
When running without -v it works fine without any error being printed, so we can ignore this.

src/algokit/cli/codespace.py
```diff
@@ -0,0 +1,148 @@
...
+    forward_ports_for_codespace,
+    is_codespace_ready,
+    list_github_codespaces,
+)
+
+logger = logging.getLogger(__name__)
+
+# https://docs.github.com/en/codespaces/setting-your-user-preferences/setting-your-timeout-period-for-github-codespaces
+MIN_GH_PORT_FORWARD_TIMEOUT_MIN = 1
+MAX_GH_PORT_FORWARD_TIMEOUT_MIN = 24
```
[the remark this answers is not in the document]
aorumbayev: @neilcampbell renamed the constants and moved them to core instead

neilcampbell (approved):

aorumbayev:
@ShoGinn codespaces support is out and available under this release version https://github.com/algorandfoundation/algokit-cli/releases/tag/v2.1.0

PR #458, opened 2024-03-19
fix: create the npm dir in the app data directory on windows, as npx needs it

neilcampbell:
Fixes issues raised here https://github.com/algorand-coding-challenges/challenge-2/pull/43


src/algokit/core/typed_client_generation.py
```diff
@@ -98,6 +99,11 @@ def __init__(self) -> None:
         if not npx_path:
             raise click.ClickException("Typescript generator requires Node.js and npx to be installed.")
 
+        # Create the npm directory inside %APPDATA% if it doesn't exist, as npx on windows needs this.
+        if is_windows():
+            npm_dir = get_app_config_dir() / "npm"
+            npm_dir.mkdir() if not npm_dir.exists() else None
```
robdmoore, line 105: This may fail due to permission errors. We need to catch and issue a console message indicating they are about to probably receive an error due to <link to github bug for npx> and they should create <folder> manually, since we couldn't for them.

neilcampbell: Yeah good idea.
We should be ok though, as it's the same root path as the algokit config directory.

neilcampbell: Fixed

PR #460, opened 2024-03-20
feat: orchestration pr feedback

aorumbayev:
## Proposed Changes

  - List command now works anywhere within the algokit workspace
  - Base template now includes default code workspace file
  - Algokit auto appends new dirs to code workspace during init 
  - Algokit prevents user from using init with workspace mode inside a workspace if init isn't executed from the root of that workspace (to prevent nested workspaces)


docs/features/init.md
```diff
@@ -42,6 +42,12 @@ type = 'workspace' # type specifying if the project is a workspace or standalone
 projects_root_path = 'projects' # path to the root folder containing all sub-projects in the workspace
 ```
 
+#### VSCode optimizations
+
+AlgoKit has a set of minor optimizations for VSCode users that are useful to be aware of:
+
+- Templates created with the `--workspace` flag automatically include a VSCode code-workspace file. New projects added to an AlgoKit workspace are also integrated into an existing VSCode workspace.
```
robdmoore, line 49: Are these two options the default?

aorumbayev: @robdmoore the --ide one has been there since the start, i noticed it wasn't documented so figured ill note it down explicitly. As for the first one its a new behaviour that this pr is adding based on feedback from @iskysun96 and discussion over it with @neilcampbell

PR #461, opened 2024-03-23
feat: LocalNet should run as an archival node so that you can access all blocks

robdmoore:
This is useful when testing.

For example, the algokit-subscriber-ts library needs to go back to block 1 and once you've been running localnet for a while and get past 1000 blocks this test now fails.

This wasn't the previous behaviour (not being archival), but I believe [this change](https://github.com/algorand/go-algorand/releases/tag/v3.22.0-stable) may have caused a change (although it's weird we had `Archival: false` and it was an archival node previously.


neilcampbell (approved):
Makes sense at face value, however also worth confirming with @aorumbayev that there wasn't a specific reason we set `Archival: false`.

aorumbayev (approved):

PR #523, opened 2024-06-19
fix: add localnet proxy to add Access-Control-Allow-Private-Network header

### LINKED ISSUE #524: Dappflow fails in Chrome
Some Chrome users have been noticing that Dappflow is failing CORS preflight requests to localnet. This is due to the following, which Google appear to be slowly rolling out. https://developer.chrome.com/blog/private-network-access-preflight.

See https://discord.com/channels/491256308461207573/1249837619336970360

To fix this the host needs to send the response header `Access-Control-Allow-Private-Network: true` when the CORS preflight request contains the header `Access-Control-Request-Private-Network: true`.

neilcampbell:
Google Chrome is slowly introducing more restrictions on sites which make calls to private network resources.
See https://developer.chrome.com/blog/private-network-access-preflight

This impacts tools like Dappflow.
This PR adds a proxy container to LocalNet and attaches the correct headers to ensure the pre flight CORS request that Chrome sends will succeed.

This PR fixes #524


robdmoore (approved):

aorumbayev:
@neilcampbell looks like poetry update urllib is needed

Issue #532, opened 2024-07-04
Trim down sensible defaults on algokit templates


aorumbayev:
### Proposal

1. .copier-answers.yml x3 -> do we really need to show them?
- - [x] - research on options behind hiding the copier answers inside .algokit file 
- - [x] - to be checked upon resolution of the item
3. envs x8 -> there are 6 env files (sc), and 2 in the frontend, could we slim them down? 
- - [x] - agree on sensible defaults to trim down the .env files, whether its removing certain network/generic .env files OR opting for extra feature in bootstrap that auto deletes the .env.template files 
4. license -> shall we nuke it? I reckon that 99% of users will do it visually by github once/if it's time to opensource the repo
- - [x] - to be checked after license is removed from all templates 
5. .editor-config x3
- - [x] - research on options around keeping a single editorconfig applied globally to all child subprojects
- - [x] - to be checked upon resolution of the item
7. .gitattributes x3 -> are they useful?
- - [x] - research on options around keeping a single gitattributes applied globally to all child subprojects
- - [x] - to be checked upon resolution of the item
9. .devcontainer.json -> should we put it behind a generator? algokit generate devcontainer
- - [x] - to be checked after devcontainer file is moved into an actionable generator supplied as part of the baseline template
11. tsconfig and tsconfig.node -> do they need to be separate files?
- - [x] - research on options to combine into one tsconfig file

aorumbayev:
@Loedn @robdmoore  open questions:

1. Do we want to:
- - a) Cut down .env file from sc contracts folder and refer to .env.localnet, .env.localnet.template, .env.testnet and .env.testnet.template files to be created by default after full init?
- - b) Expand bootstrap to auto delete .env.template files after it creates a base .env from that template?
- - c) Ignore a,b and instead have a generic generator thats invocable from the root of individual projects (would require 1 unique smart contract specific env generator as it needs to be aware of deploy conventions) and generic generator on frontend OR we only introduce that generator on smart contracts projects given that its a single .env and .env.template on frontend that is fairly conventional setup for frontend frameworks already

robdmoore:
Single gitattributes in git root will work, no need to research that.

robdmoore:
(but if you do single project mode rather than workspace you should keep the gitattributes)

robdmoore:
[...]

PR #533, opened 2024-07-05
feat: adding default algorand network configs to use when no .env.{network} found

aorumbayev:
https://github.com/algorandfoundation/algokit-cli/issues/532

## Proposed changes

Adding defaults to use if user is invoking algokit project deploy {mainnet|localnet|testnet} and no .env.{network_name} exists. This should not prevent user from deploying with defaults (and in case of testnet|mainnet it will use free public infrastructure - algonode). But the main reason is to reduce noisy amount of .env files in smart contract based algokit templates.


src/algokit/core/project/deploy.py
```diff
@@ -4,13 +4,55 @@
...
+    _KnownEnvironments.LOCALNET: {
+        # this file should contain environment variables specific to algokit localnet
+        "ALGOD_TOKEN": str(DEFAULT_ALGOD_TOKEN),
+        "ALGOD_SERVER": str(DEFAULT_ALGOD_SERVER),
+        "ALGOD_PORT": str(DEFAULT_ALGOD_PORT),
+        "INDEXER_TOKEN": str(DEFAULT_INDEXER_TOKEN),
+        "INDEXER_SERVER": str(DEFAULT_INDEXER_SERVER),
+        "INDEXER_PORT": str(DEFAULT_INDEXER_PORT),
+    },
+    _KnownEnvironments.MAINNET: {
```
neilcampbell, line 45: Do we need ports for the mainnet and testnet configs?

src/algokit/core/project/deploy.py
```diff
@@ -4,13 +4,55 @@
...
 
+class _KnownEnvironments:
+    LOCALNET = "localnet"
+    MAINNET = "mainnet"
+    TESTNET = "testnet"
+
+
+_mainnet_algod_config = get_algonode_config("mainnet", config="algod", token=DEFAULT_ALGOD_TOKEN)
+_testnet_algod_config = get_algonode_config("testnet", config="algod", token=DEFAULT_ALGOD_TOKEN)
+_mainnet_indexer_config = get_algonode_config("mainnet", config="indexer", token=DEFAULT_INDEXER_TOKEN)
```
neilcampbell, line 31: AFAIK no token is needed for algonode.

robdmoore: +1

src/algokit/core/project/deploy.py
robdmoore: +1 - should be 443 right?

aorumbayev: @robdmoore @neilcampbell i used the values from localnet config from .env.template files we got on smart contract templates. The values are as follows:

ALGOD_TOKEN=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
ALGOD_SERVER=http://localhost
ALGOD_PORT=4001
INDEXER_TOKEN=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
INDEXER_SERVER=http://localhost
INDEXER_PORT=8980

I am pretty sure ports are needed but only on localnet, there is a reason why we got it this way on the template repos. Will double check the behaviour where ports are utilized specifically in localnet cases, i think there is some methods in algokit-utils that rely on existence of those vars in localnet scenarios.

PR #550, opened 2024-08-02
fix: build binaries with the correct version information

neilcampbell:
When building the binaries we were building based on the old version metadata.
This PR fetches the code based on the release tag ref, so that we always build with the correct metadata, as `actions/checkout` will fetch based on the commit hash which triggered the build (not using the main ref), when no ref is supplied.

### DIFF
--- a/.github/workflows/build-binaries.yaml
+++ b/.github/workflows/build-binaries.yaml
@@ -9,6 +9,9 @@ on:
       python_version:
         required: true
         type: string
+      release_version:
+        required: false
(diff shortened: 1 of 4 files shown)


aorumbayev (approved):
Verified in a forked repo against dummy action run [[fix-binary-cli-version](https://github.com/algorandfoundation/algokit-cli/tree/fix-binary-cli-version)](https://github.com/aorumbayev/algokit-cli/actions/runs/10213704219)

PR #557, opened 2024-08-18
chore: explicit sequential execution; extra args param for run/deploy commands

aorumbayev:
## Proposed Changes

  - Adds ability to pass extra args to deploy/run commands (has to be separated via `--`) . This can be reused in templates to allow deploy/building individual contracts/projects by name as extra arg.
  - Adds explicit flags (concurrent mode by default) on workspace commands where user can explicitly run it in sequential mode (without specifying order in toml file). Unblocks edge cases reported by @CiottiGiorgio  where multiple commands can't be executed concurrently and user doesn't have a preference on which exact order of execution is it they just need it to run sequentially in any order. 
- Updates docs with minor sections on new cli param
- Adding a few more comprehensive tests for binary execution mode (which implicitly verifies latest extra args behaviour)

## TODO

- [x] - Snapshot tests


docs/features/project/run.md
```diff
@@ -113,27 +127,46 @@ Executing `algokit project run hello` from the root of the workspace will concur
 
 Executing `algokit project run hello` from the root of `project_(a|b)` will execute `echo hello` in the `project_(a|b)` directory.
 
-### Controlling order of execution
+### Controlling Execution Order
 
-To control order of execution, simply define the order for a particular command as follows:
+Customize command execution in workspaces for precise control:
```
neilcampbell, line 132: I think we split the order for order from the concurrency docs. They are different things.

src/algokit/cli/project/run.py
```diff
@@ -118,6 +122,15 @@ def base_command(
                 default=None,
                 help="Limit execution to specific project types if executing from workspace. (Optional)",
             )(command)
+            command = click.option(
```
neilcampbell, line 125: I'm wondering if we invert this to `Execute workspace commands sequentially`, that's the feature we're trying to unlock and it also means it's default false as concurrent is the default.

docs/features/project/run.md
aorumbayev: @neilcampbell pushed some changes

src/algokit/cli/project/run.py
aorumbayev: good point, inverted

docs/features/project/run.md
neilcampbell: @aorumbayev I can't see the changes. Did you push them.

aorumbayev: @neilcampbell should be visible now

neilcampbell (approved):

PR #563, opened 2024-09-03
chore: updating localnet config update log message for --no-update case

aorumbayev:
## Proposed Changes

  - Minor patch to change wording when for log message informing on update of the localnet


aorumbayev:
@neilcampbell reverted the changes, the latest is just tweak in the message

neilcampbell (approved):

PR #565, opened 2024-09-09
chore: bumping cryptography (pip-audit warning)

aorumbayev:
Currently blocked due to https://github.com/auth0/auth0-python/issues/627.


aorumbayev:
@neilcampbell we can either:
1. Move the related abstractions (its just 2 classes that we rely on in this package around JWT validation) - we can remove dependency on auth0 completely (it was known to cause similar issues for at least twice over past year) 
2. Wait for this to be addressed on auth0 package, looks like issue 627 has a lot of folks pushing for resolution 
3. Ignore the warning on pip audit side and apply 2 as soon as the fix is available.

What do you think ?

neilcampbell:
@aorumbayev If we can remove the dependency easily, then maybe that's a good way forward.

aorumbayev:
@neilcampbell moved auth0 specific depenendcies to _vendor folder inside core, header include ref to file and ref commit. Let me know what you think

neilcampbell:
@aorumbayev Nice. Assuming you've checked the testnet dispenser features are all working as expected?

neilcampbell (approved):

PR #566, opened 2024-09-10
chore: removing aiohttp dependency (given that auth0 is now vendored)

aorumbayev:
Tiny follow up to previous pr removing auth0 package, realized that aiohttp was only brought in due to dependency on auth0 package, we no longer need it either so removing it


neilcampbell (approved):
Nice!

Issue #568, opened 2024-09-13
Ability to pass dev mode on/off as part of named localnet start, ability to specify custom path to named localnets 


aorumbayev:
## Problem

There is a certain amount of friction and requirement on internals to understand where to place named localnet when executing in CI environments. 

A user feedback request suggests ability to allow configuring dev mode for algod directly via flag in algokit CLI and a simpler flow to execute named localnets in CI environments

## Solution

- Introducing an extra flag to specify a path to named localnet to load - allowing user to save the required custom localnets in version control to be easily invoked within CI environment. Whether this should be done via direct cli argument or an env var or both is to be evaluated
- Introducing a flag specificalyl for the dev mode on/off on algod as part of executing a named localnet. The main argument for is the fact this is a very common configuration option that can be useful both locally and within CI - to be evaluated.

aorumbayev:
@pbennet this is now available in 2.4.0 , refer to new sections in localnet command group docs for extra references, https://github.com/algorandfoundation/algokit-cli/blob/main/docs/features/localnet.md#controlling-algod-developer-mode

PR #569, opened 2024-09-14
feat: adding flags to enable dev mode; flags for custom path to localnet config

aorumbayev:
Fixes https://github.com/algorandfoundation/algokit-cli/issues/568

## Proposed Changes

  - Adds -dev/--no-dev flag for controlling devmode on algod container
  - Adds option to specify custom path to compose config directory


src/algokit/cli/localnet.py
```diff
@@ -126,7 +128,26 @@ def config_command(*, engine: str | None, force: bool) -> None:
     "AlgoKit will not manage the configuration of named LocalNet instances, "
     "allowing developers to configure it in any way they need.",
 )
-def start_localnet(name: str | None) -> None:
+@click.option(
```
neilcampbell, line 131: Do we want to ensure that a user doesn't supply both `name` and `config_path`?

src/algokit/core/sandbox.py
```diff
@@ -46,9 +46,9 @@ def get_min_compose_version() -> str:
 
 
 class ComposeSandbox:
```
neilcampbell, line 48: There is logic to detect if any configuration is out of date. Do we need to change that to ignore the value of dev mode?

src/algokit/core/sandbox.py
```diff
@@ -184,6 +195,21 @@ def up(self) -> None:
         else:
             logger.warning("AlgoKit LocalNet failed to return a successful health check")
 
+    def recreate_container(self, container_name: str) -> None:
```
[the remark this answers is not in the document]
aorumbayev: Restored to regular reset, recreate would result in faulty state

src/algokit/cli/localnet.py
aorumbayev: Why do you think doing both is not an option?

docs/features/localnet.md
```diff
@@ -158,5 +187,6 @@ Running an interactive session ensures that you have control over the lifecycle
 - `--force`, `-f`: Force deletes stale codespaces and skips confirmation prompts. Defaults to explicitly prompting for confirmation.
 
 For more details about managing LocalNet in GitHub Codespaces, please refer to the [AlgoKit CLI reference documentation](../cli/index.md#codespace).
+U
```
neilcampbell, line 190:
```suggestion
U
```

PR #599, opened 2024-12-24
fix: handle client generation in a dir containing multiple app spec types

### LINKED ISSUE #598: `algokit generate client` should detect the most applicable app specs in a directory
From discord https://discord.com/channels/491256308461207573/1319528059090501743/1319528059090501743

When running `algokit generate client ../../built_contracts/v3 --language typescript  --output  ./clients/{contract_name}Client.ts` and the directory contains both ARC56 and ARC32 app specs for a single app, the CLI should detect the most applicable (ARC56) and only run the client generation once. Currently it will generate the client using the ARC32 file, then generate using the ARC56 file and overwrite the previously generated client.

neilcampbell:
Resolves #598


src/algokit/core/typed_client_generation.py
```diff
@@ -88,6 +102,43 @@ def find_generate_command(self, version: str | None) -> list[str]: ...
     def format_contract_name(self, contract_name: str) -> str:
         return contract_name
 
+    def generate_all(
+        self,
+        app_spec_path_or_dir: Path,
+        output_path_pattern: str | None,
+        *,
+        raise_on_failure: bool,  # TODO: NC - Maybe we should return the error instead?
```
aorumbayev, line 110: Perhaps another option could be to remove raise_on_failure and add dataclass to represent result from generate_all response that will contain success and failure, something like
```@dataclass
    class GenerateAllResult:
        successful_generations: list[tuple[Path, Path]]  # List of (app_spec, output_path) that succeeded
        failed_generations: list[tuple[Path, str]]  # List of (app_spec, error_message) that failed
        
        @property
        def has_failures(self) -> bool:
            """Returns True if any generations failed."""
            return len(self.failed_generations) > 0
 ```
 
 and then the generate.py would be taking care of deciding when to raise an exception

aorumbayev (approved):

src/algokit/core/typed_client_generation.py
neilcampbell: Yeah I was tossing up that option and forgot to remove the comment.
This approach would be behaviourally different though, as the exception is thrown on the first failure. It's tied into the `fail_fast` option in the `link` command, so we actually want to abort the generation on failure.

neilcampbell: It's actually a bit of a weird one, as it only handles path resolution failure, which is what the previous code was doing. I'm going to keep the boolean and rename to be more specific.

PR #601, opened 2025-01-08
fix: removed mentions to dappflow from ADR document.

CiottiGiorgio:


neilcampbell:
Thanks @CiottiGiorgio. It reminds me that we also need to update the readme image. I'll send another PR for that.

neilcampbell (approved):

CiottiGiorgio:
Thanks

PR #604, opened 2025-01-08
chore: adding conditional to skip codesigning on PR triggers when building binaries

aorumbayev:
The main gist is to further refine access to secrets to ensure none are accessed on PR triggers. Codesigning shall only run on production releases on main branch OR cron triggered checks performed weekly (8am UTC mondays) on main branch. 

## Proposed Changes
  - Add a new "with_codesign" parameter to macOS and Windows build actions to control whether binaries should be signed.
  - Update the build-binaries workflow to determine code-signing eligibility (via a new "signing" step) based on branch, PR status, and release conditions.
  - Update the macOS composite action to set "APPLE_CERT_ID" and "APPLE_BUNDLE_ID" conditionally, falling back to a beta bundle ID when "production_release" is false.
  - Replace references to "production_release" in the Windows build action with "with_codesign" for consistency.
  - Modify pyproject.toml to allow conditional code signing by passing in environment variables, ensuring more flexible signing logic for macOS binaries.

### DIFF
--- a/.github/actions/build-binaries/macos/action.yaml
+++ b/.github/actions/build-binaries/macos/action.yaml
@@ -9,6 +9,9 @@ inputs:
     required: true
   version:
     description: "The version to use for this artifact"
+  with_codesign:
+    description: "Flag to determine if we should sign the binary"
+    required: true
   apple_team_id:
     description: "The Apple Team ID"
     required: true
@@ -20,10 +23,10 @@ inputs:
     required: true
   apple_notary_user:
     description: "The Apple user to notarise the package"
-    require: true
+    required: true
   apple_notary_password:
     description: "The Apple password to notarise the package"
-    require: true
+    required: true
 
 runs:
   using: "composite"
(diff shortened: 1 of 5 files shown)


.github/actions/build-binaries/macos/action.yaml
```diff
@@ -20,23 +23,23 @@ inputs:
     required: true
   apple_notary_user:
     description: "The Apple user to notarise the package"
-    require: true
+    required: true
```
lempira, line 26: Good catch. all of these "required" wont fail because you fall back to '' when code_sign=false, right?

.github/workflows/build-binaries.yaml
```diff
@@ -24,6 +24,21 @@ jobs:
...
+          # 2. Main branch CRON triggered events
+          if [[ "${{ github.event_name }}" != "pull_request" && \
+                "${{ github.ref_name }}" == "main" && \
+                ("${{ inputs.production_release }}" == "true" || "${{ github.event_name }}" == "schedule") ]]; then
+            echo "allowed=true" >> $GITHUB_OUTPUT
+          else
+            echo "allowed=false" >> $GITHUB_OUTPUT
+          fi
+        shell: bash
+
```
lempira, line 41: I wonder if this could be made more succinct. Something like:
`echo "allowed=${{ github.event_name != 'pull_request' && github.ref_name == 'main' && (inputs.production_release == 'true' || github.event_name == 'schedule') }}" >> $GITHUB_OUTPUT`

and then put the comments in the yaml. Just a thought, but up to you if you want to change.

lempira (approved):

.github/actions/build-binaries/macos/action.yaml
aorumbayev: yes exactly, when code signing is false it will still pass these arguments as empty strings

aorumbayev:
@neilcampbell @lempira - instead of forking and running whole e2e i ended up force setting codesigning flag to true to test the codesigning parts, you can see the run passing here https://github.com/algorandfoundation/algokit-cli/actions/runs/12680599041/job/35342740513. The last commit reverts the temporary forced true flag.

.github/workflows/build-binaries.yaml
aorumbayev: I hit some issues with this approach since value ends up being a bool, we used a string literal for production_release parameter so i ended up reusing that approach instead. An explicit `true` `false` strings seem to be easier to deal with thats why the script is a bit more verbose

PR #605, opened 2025-01-13
feat: npm runs `ci` instead on `install` when CI flag is present or passed explicitly`

aorumbayev:
# Proposed changes

- Adds extra param to npm sub command in `bootstrap` group to run `npm ci` when CI is present as env var and set to true or passed via `--ci` flag (or `--non-interactive/--ci` in `all` command)
- If package lock is not present during npm ci execution it will present and error.

### DIFF
--- a/src/algokit/cli/project/bootstrap.py
+++ b/src/algokit/cli/project/bootstrap.py
@@ -43,7 +43,7 @@ def bootstrap_group(ctx: click.Context, *, force: bool) -> None:
 )
 @click.option(
     "--interactive/--non-interactive",
-    " /--ci",  # this aliases --non-interactive to --ci
+    "--no-ci/--ci",  # this aliases --non-interactive to --ci and --interactive to --no-ci
     default=lambda: "CI" not in os.environ,
     help="Enable/disable interactive prompts. If the CI environment variable is set, defaults to non-interactive",
 )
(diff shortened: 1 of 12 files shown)


src/algokit/cli/project/bootstrap.py
```diff
@@ -102,5 +102,11 @@ def poetry() -> None:
 @bootstrap_group.command(
     "npm", short_help="Runs `npm install` in the current working directory to install Node.js dependencies."
 )
-def npm() -> None:
-    bootstrap_npm(Path.cwd())
+@click.option(
```
neilcampbell, line 105: Does that change apply to `all` as well?

src/algokit/cli/project/bootstrap.py
```diff
@@ -102,5 +102,11 @@ def poetry() -> None:
...
+    "--ci",
+    is_flag=True,
+    default=lambda: "CI" in os.environ,
```
neilcampbell, line 108: Confirming this is automatically set on a GitHub actions environment?

neilcampbell (approved):

src/algokit/cli/project/bootstrap.py
aorumbayev: Yes it is, see https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

src/algokit/cli/project/bootstrap.py
aorumbayev: `all` already has --ci, --interactive/--no-interactive options where (--ci is equivalent to --no-interactive).

PR #611, opened 2025-02-10
chore: algokit utils v3 migration

aorumbayev:
## Proposed Changes

  - Migrating codebase to rely on algokit-utils v3
  
## Further notes

- Update to production release after prod utils v3 is available on pypi
- @lempira no major bumps in semantic version needed, even though it affects the cli features, cli itself isn't meant to be consumed as a dependency to be used programmatically, hence from that standpoint no major bumps are needed in the cli.
- @lempira please note, the portability tests are failing since they run against main branch of the python template, the ci should be retriggered and go green after prod PR updating to support utils-v3 is merged to main https://github.com/algorandfoundation/algokit-python-template/pull/94


neilcampbell (approved):

.github/workflows/check-python.yaml
```diff
@@ -24,12 +24,17 @@ jobs:
...
           # audit non dev dependencies, no exclusions
-          poetry export --without=dev > requirements.txt && poetry run pip-audit -r requirements.txt
-
+          poetry export --without=dev > requirements.txt
+          poetry run pip-audit -r requirements.txt  --ignore-vuln 'GHSA-79v4-65xg-pq4g'
           # audit all dependencies, with exclusions.
           # If a vulnerability is found in a dev dependency without an available fix,
           # it can be temporarily ignored by adding --ignore-vuln e.g.
-          poetry run pip-audit
+          # TODO: decide on `GHSA-79v4-65xg-pq4g`, see https://osv.dev/vulnerability/GHSA-79v4-65xg-pq4g
```
aorumbayev, line 32: @lempira i ended up ignoring the vulnerability for now, pip-audit has been a contender for removal in algokit-cli for a while because of the distribution model over pipx, in such cases the contents of the lockfile don't matter for the end user - while it allows us to track the lockfile on the repo, pipx will ignore the lockfile during installation of the package as a pure. See https://github.com/pdm-project/pdm/discussions/2450. Given this being out of scope for the intends of this pr, i'd suggest tackling this as a separate fix. If you'd decide to keep using pip-audit, a bump in cryptography to 44.0.1 should be sufficient but i noticed it causing additional issues with binary executable because it seems like the cryptograph 44.0.1 wheels were build on a machine that with glibc of version higher than GLIBC_2.33. This also warranted this being a bit more complex than just a version bump so a proper dedicated item to tackle it is worth having.

lempira (approved):

PR #612, opened 2025-03-05
feat: puya-ts support under `compile` command group

aorumbayev:
## Proposed Changes

  - Adds support for puya-ts, relying on npx and its internal module resolution behavior 

## TODO

- [x] Await for finalization of changes in regards to puya binaries before merging
- [x] Clarify on whether explicit `build` is required, or whether it should be removed to align with puyapy cli interface
- [x] Await for resolution on bundling stubs with the puya ts package


neilcampbell (commented):
We need to probably also update the compile feature docs.

aorumbayev:
@neilcampbell added extra tests and reflected latest change on puya-ts that completely aligns it with puyapy cli entrypoint

neilcampbell (approved):

PR #618, opened 2025-03-21
chore: fix audit issue

neilcampbell:
Resolves the audit issues against the `cryptography` package.

The Auth0 package has relaxed their version constraints in https://github.com/auth0/auth0-python/pull/660, which indicates that the latest version is compatible with the latest `cryptography` and `pyjwt` packages. I diffed the latest version of the Auth0 `token_verifier.py` against the copy we have vendored and nothing meaningful has changed.

Additionally I have tested the dispenser features to confirm everything functions correctly.

### DIFF
--- a/.github/workflows/check-python.yaml
+++ b/.github/workflows/check-python.yaml
@@ -25,16 +25,10 @@ jobs:
         run: |
           # audit non dev dependencies, no exclusions
           poetry export --without=dev > requirements.txt
-          poetry run pip-audit -r requirements.txt  --ignore-vuln 'GHSA-79v4-65xg-pq4g'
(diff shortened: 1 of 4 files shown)


neilcampbell:
This PR requires more effort to action, as `cryptography` 44+ ships with `manylinux_2_34_*` wheels. The ubuntu 20 agent we use for building binaries has a lower glibc version available and fails. A better approach is to refactor the binary build pipeline to use the method recently established in the `puya` repo.

neilcampbell:
Now that we're building on ubuntu 22 agents, this PR can be re-opened without any changes to the build configuration.

lempira (approved):

PR #620, opened 2025-03-25
fix: resolve project run issue by forcing utf-8 encoding on windows

neilcampbell:
Fixes https://github.com/algorandfoundation/algokit-client-generator-ts/issues/154
By forcing the `Popen` used in `run` to use utf-8 encoding we can bypass Windows using the default cp1252 encoding, which causes strange issues.

Additionally I have update a few other places to force utf-8 encoding.

After merging (before release), I'll do some additional smoke tests of the various code paths.

### DIFF
--- a/src/algokit/cli/project/link.py
+++ b/src/algokit/cli/project/link.py
@@ -239,5 +239,5 @@ def link_command(
             version=version,
         )
 
-        logger.info(f"✅ {iteration}/{total}: Finished processing {contract_project.project_name}")
+        logger.info(f"{iteration}/{total}: Finished processing {contract_project.project_name}")
         iteration += 1
--- a/src/algokit/cli/dispenser.py
+++ b/src/algokit/cli/dispenser.py
@@ -61,7 +61,7 @@ def _handle_ci_token(output_mode: str, output_filename: str, token_data: dict) -
             "If needed, clear your terminal history after copying the token!"
         )
     else:
-        with Path.open(Path(output_filename), "w") as token_file:
+        with Path.open(Path(output_filename), mode="w", encoding="utf-8") as token_file:
             token_file.write(token_data["access_token"])
         logger.warning(
             f"Your CI access token has been saved to `{output_filename}`.\n"
--- a/src/algokit/cli/tasks/analyze.py
+++ b/src/algokit/cli/tasks/analyze.py
@@ -223,7 +223,7 @@ def analyze(  # noqa: PLR0913, C901
                 if has_diff:
                     raise click.exceptions.Exit(1)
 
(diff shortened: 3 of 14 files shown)


aorumbayev (approved):

src/algokit/cli/project/link.py
```diff
@@ -239,5 +239,5 @@ def link_command(
             version=version,
         )
 
-        logger.info(f"✅ {iteration}/{total}: Finished processing {contract_project.project_name}")
+        logger.info(f"{iteration}/{total}: Finished processing {contract_project.project_name}")
```
lempira, line 242: Why was the emoji removed? Does this have to do with utf-8 encoding?

lempira (commented):
Is it possible/favorable to create an extension of Path that encodes utf-8 so you don't have to define it everywhere? Something like 

```python
from pathlib import Path

class Utf8Path(Path):
    def read_text(self, encoding="utf-8", **kwargs):
        return super().read_text(encoding=encoding, **kwargs)
        
    def write_text(self, content, encoding="utf-8", **kwargs):
        return super().write_text(content, encoding=encoding, **kwargs)
    
    def open(self, mode="r", encoding="utf-8", **kwargs):
        if "b" not in mode:
            return super().open(mode=mode, encoding=encoding, **kwargs)
        return super().open(mode=mode, **kwargs)

```

src/algokit/cli/project/link.py
neilcampbell: I removed it because the link command is usually called via `algokit project run build` which calls `npm run build` which calls `npm run generate:app-clients` which finally calls `algokit project link` in the frontend template.

The user will only see this output when running in verbose mode if run in this way. It looked a bit weird in the verbose output, so removed it as I don't think it's really needed.

neilcampbell:
> Is it possible/favorable to create an extension of Path that encodes utf-8 so you don't have to define it everywhere? Something like
> 
> ```python
> from pathlib import Path
> 
> class Utf8Path(Path):
>     def read_text(self, encoding="utf-8", **kwargs):
>         return super().read_text(encoding=encoding, **kwargs)
>         
>     def write_text(self, content, encoding="utf-8", **kwargs):
>         return super().write_text(content, encoding=encoding, **kwargs)
>     
>     def open(self, mode="r", encoding="utf-8", **kwargs):
>         if "b" not in mode:
>             return super().open(mode=mode, encoding=encoding, **kwargs)
>         return super().open(mode=mode, **kwargs)
> ```

Possibly. I'm not sure it's that much better, as we'd still need to wrap the Path with Utf8Path everywhere.
I was trying to keep the change set small here.

lempira (approved):

PR #623, opened 2025-03-27
fix: raise min required copier version to 3.6.0

aorumbayev:
## Proposed changes

- Fixes issue reported by @d13co where upgrading from older algokit cli versions results in copier dependency not being bumped to latest -> resulting in failing pydantic errors during init against fullstack template 
- Minor tweaks to fix failing tests for ts compile command (ensuring it sticks to beta versions for now)
- Added extra verbose method to doctor command to optionally output direct algokit package dependencies used -> due to pipx not relying on lock files during install we can't quickly diagnose user reported errors knowing exact package dependencies. This would allow slightly simplifying such scenarios allowing quicker way to retrieve exact versions (note that it only concerns cases when algokit is installed as a wheel, obtaining versions of packages is not trivial when inside pyinstaller binary so i skipped adding more complexity there)


src/algokit/cli/doctor.py
```diff
@@ -35,7 +36,15 @@
     is_flag=True,
     default=False,
 )
-def doctor_command(*, copy_to_clipboard: bool) -> None:
+@click.option(
```
neilcampbell, line 39: AlgoKit already support a verbose flag at the root command, so it would be possible for a user to run `algokit -v doctor -v` which feels a little strange. Could/Should we detect the `-v` flag at the `algokit` level and activate instead?

@lempira was also talking about potentially leveraging doctor to expose project package information when run inside an AlgoKit project. We probably want to consider how that'd impact a change like this.

I'm wondering if this feature should be extracted into a seperate PR outside the fix, so we can consider a few other facets?

tests/compile/test_typescript.py
```diff
@@ -63,12 +63,13 @@ def typescript_test_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
...
 
     # Create package.json with required dependencies
+    # TODO: update to use latest versions once they are released out of beta
     package_json_content = """{
         "name": "algokit-test",
         "version": "1.0.0",
         "dependencies": {
-            "@algorandfoundation/puya-ts": "^1.0.0-beta.48",
-            "@algorandfoundation/algorand-typescript": "^1.0.0-beta.25"
+            "@algorandfoundation/puya-ts": "~1.0.0-beta.48 <1.0.0",
```
neilcampbell, line 71: Why was this needed?
We can also probably bump to the latest beta versions.

src/algokit/cli/doctor.py
aorumbayev: As discussed on a call, this doesnt add any new commands groups options or arguments to doctor (i will reuse global verbose flag instead) - all it changes is appending extra info that can be super helpful when debugging user queries. If @lempira would be adding new functionality to doctor it would require changes regardless + would potentially be handled via subcommand group, so i don't this will complicate this or overlap with potential new behaviour that would be introduces

neilcampbell (approved):

PR #629, opened 2025-04-09
chore: bumped ubunto build binary version

lempira:
Bumped the ubuntu os version in the build_binaries workflow to 22.04 since 20.04 is being deprecated.


neilcampbell (approved):
Also once this change has gone in, we can re-open this PR https://github.com/algorandfoundation/algokit-cli/pull/618

PR #634, opened 2025-04-29
chore: updated relevant to packages in order to update h11

lempira:
Updates pytest, pytest-httpx, and h11. 

Had to updates several snapshot and update some test for new API of pytest-httpx. 

~~This PR is in draft and will be released once this PR is merged:~~

~~https://github.com/algorandfoundation/algokit-utils-py/pull/157~~

Algokit-utils-py PR has been merged

This update included many unexpected update due to:
- the bump in httpx now fails on all HTTP calls that are not mocked. This caught an error where a test was [failing for the wrong reason](https://github.com/algorandfoundation/algokit-cli/pull/634/commits/d17238532f2dfbabe28188e4f4675454612dfa35).
- The minor bump in copier actually had some breaking changes
  - [v9.7.0 now requires to pass in a context to the `Question`](https://github.com/algorandfoundation/algokit-cli/pull/634/commits/3bb10295be17d87077186cd560de1b16834079fc)
- Was getting deprecation errors for importing from modules in copier. For ex. `from copier.main import Worker` was getting a warning of `DeprecationWarning: Importing from `copier.main` is deprecated. This module is intended for internal use only and will become inaccessible in the future.` [The change was `from copier._main import Worker`](https://github.com/algorandfoundation/algokit-cli/pull/634/commits/ed4aae57cc8b6a7475e78542fec6f241eb2b9eb7)


pyproject.toml
```diff
@@ -24,21 +24,21 @@ keyring = "25.2.1"
 # this version has been tested to work with the vendored file
 pyjwt = "^2.10.1" 
 cryptography = "^44.0.2" # pyjwt has a weak dependency on cryptography and explicitly requires it in the vendored file, hence the lock
-algokit-utils = "^3.0.0"
+algokit-utils = "4.0.1b3"
```
neilcampbell, line 27: Before merging this change we should release algokit-utils and update to the latest version.

neilcampbell (approved):

pyproject.toml
lempira: Released 4.0.1 of utils-py and updated the pyproject and lock files. I also pinned the copier version as @aorumbayev suggested. Since the minor update to copier had breaking changes, we can pin the version and manage through dependabot.

neilcampbell: Looks good to me.

PR #702, opened 2025-12-06
feat: reduce localnet start/reset time

neilcampbell:
Each time LocalNet is started or reset, a number of http calls would be made to the docker hub to check if any new docker image versions are available. New versions are generally released infrequently, so most of the time this check is not needed and slows the start/reset time of LocalNet.

This PR reduces the frequency of calls to check for new docker image versions, so that start/reset time is faster.
Based on some rudimentary tests, the following results were observed:

**Starting without existing containers** - `algokit localnet start`
Baseline: 10.51s (average)
Without version check: 7.82s (average)
≈ 25% improvement

**Starting with existing containers** - `algokit localnet start`
Baseline: 3.767s (average)
Without version check: 1.845s (average)
≈ 51% improvement

**Resetting** - `algokit localnet reset`
Baseline: 13.674s (average)
Without version check: 11.304s (average)
≈ 17% improvement

### DIFF
--- a/docs/cli/index.md
+++ b/docs/cli/index.md
@@ -127,30 +127,34 @@
     - [Options](#options-17)
     - [--update, --no-update](#--update---no-update)
     - [-P, --config-dir ](#-p---config-dir-)
+    - [--check](#--check)
     - [start](#start)
     - [Options](#options-18)
     - [-n, --name ](#-n---name--1)
     - [-P, --config-dir ](#-p---config-dir--1)
     - [-d, --dev, --no-dev](#-d---dev---no-dev)
     - [--force](#--force)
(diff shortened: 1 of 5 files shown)


lempira (commented):
Nice improvement!
It looks like `localnet --update` will override the image cache and pull the lates but there doesn't seem to be a way to check if the cache is outdated within the TTL (a week). What about adding another localnet command like `algokit localnet check` that will just run the check and compare it against the cache? Alternatively, there could be a `--check` flag that will perform the check on that command. So `algokit localnet start --check` will start localnet but warn if there are new images. It would also work on `localnet status --check`

lempira (approved):

PR #709, opened 2026-02-05
docs: discussion doc on alternative distribution mechanisms for the cli

aorumbayev:
ADR for discussion on replacing and improving current distribution mechanisms

### DIFF
--- a/docs/architecture-decisions/2026-02-05_simplify-distribution-options.md
+++ b/docs/architecture-decisions/2026-02-05_simplify-distribution-options.md
@@ -1,0 +1,126 @@
+# Simplify AlgoKit CLI distribution (2026)
+
+- **Status:** Draft for discussion
+- **Date:** 2026-02-05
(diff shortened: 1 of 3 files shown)


joe-p:
Option A sounds good with me. How do we currently handle Puya? It would be nice to align binary distribution.

lempira:
Option A sounds good to me as well. I like the idea of having uv handling the Python distributions. I'm still a little bit unsure about all the migration paths, but I am sure we can figure that out during the implementation.

mrcointreau:
If I understood correctly the option A would allow us to replace the binary shipping with a bootstrap script that 1. installs uv, 2. installs python using uv and 3. installs algokit using uv, right? In that case this would allow us to get rid of PyInstaller, code signing, brew, snap and winget at once, I mean, to me this looks like a no-brainer (but I might be missing more than something)

aorumbayev:
@lempira i pushed a commit updating that we agreed on option A, let me know if i can merge

lempira (approved):
We decided on option A.

PR #712, opened 2026-03-03
refactor: migrate to uv-first packaging and distribution

aorumbayev:
## Why
Move AlgoKit CLI to a uv-first distribution model and remove implicit tooling behavior that was hard to reason about.

## What Changed
- Migrate Python packaging from Poetry metadata/lockfiles to uv-compatible project configuration.
- Replace CI setup paths from Poetry to uv across workflows and composite actions.
- Update runtime/bootstrap/tooling code paths for uv-first behavior.
- Keep legacy installation methods documented, but make uv the primary install path.


aorumbayev:
@lempira , github is down atm, tested ci in act - will retrigger once gh is up to ensure ci is green before publishing pr
CleanShot 2026-03-03 at 21 02 32@2x

src/algokit/core/compilers/python.py
```diff
@@ -24,24 +24,19 @@ def _find_puyapy_command_at_version(version: str) -> list[str]:
             ):
                 return puyapy_command
 
-    pipx_command = find_valid_pipx_command(
-        "Unable to find pipx install so that the `PuyaPy` compiler can be run; "
-        "please install pipx via https://pypa.github.io/pipx/ "
+    pipx_command = find_valid_x_command(
```
lempira: nit: should the var name be renamed to not use `pipx` but something more general like `valid_puyapy_command`? Also can the `_x_` be renamed to something more explicit like `find_valid_tool_runner_command`?

.github/assets/banner.svg
lempira, line 1: Thanks for suggesting a new banner. I think we have to check with the artifacts that marketing has, but we can merge this and follow up in another PR.

scripts/install/install.ps1
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function Install-UV {
+    if (Get-Command uv -ErrorAction SilentlyContinue) {
+        $uvVersion = (uv --version 2>$null) -replace 'uv ', '' -replace '\n', ''
+        Write-Log "uv found: uv $uvVersion"
+        return
+    }
+
+    Write-Log "Installing uv..."
```
lempira, line 35: The bash script has a size checker to check for corrupted files. Should this be added to the ps1 file?

```
[[ $(wc -c < "$temp_file") -ge 1000 ]] \
        || die "uv installer appears corrupted (too small)."
```

src/algokit/cli/__init__.py
```diff
@@ -67,3 +78,30 @@ def algokit(*, skip_version_check: bool) -> None:
...
 algokit.add_command(compile_group)
 algokit.add_command(project_group)
+
+
+def _check_binary_deprecation() -> None:
+    if not is_binary_mode():
+        return
+
+    import logging
+    import platform
```
lempira: Can these be moved to the top of the file?

src/algokit/core/project/run.py
```diff
@@ -19,6 +20,17 @@
...
 
+def _normalize_legacy_runner_command(command: list[str]) -> list[str]:
+    """Normalize legacy runner commands to supported equivalents."""
+    if command[:2] == ["poetry", "run"] and command[2:]:
+        if shutil.which("poetry"):
+            return command
+        if shutil.which("uv"):
+            logger.debug("poetry not found, rewriting 'poetry run' -> 'uv run' for: %s", " ".join(command))
+            return ["uv", "run", *command[2:]]
+    return command
```
lempira, line 35: If the user runs `poetry run pytest` poetry and uv are both not found, wouldn't the command fail? Should that be caught here or will a `command not found` eventually be displayed to the user?

README.md
```diff
@@ -85,99 +85,89 @@ This is an open source project managed by the Algorand Foundation. See the [cont
...
-> If you have previously installed AlgoKit using `pipx` and would like to switch to a different installation method, please ensure that
-> you first uninstall the existing version by running `pipx uninstall algokit`. Once uninstalled, you can follow the installation instructions for your preferred platform.
+## Quick Install (Recommended)
+
+AlgoKit is installed via [uv](https://docs.astral.sh/uv/), which handles Python and dependency management automatically.
+
+### macOS / Linux
+
+```shell
+curl -fsSL https://raw.githubusercontent.com/algorandfoundation/algokit-cli/main/scripts/install/install.sh | bash
```
lempira: Can we update this to the URL for the decoupling branch, and then we can update it for main before we merge?

lempira (approved):
Looks good. Just a couple of nit comments.

PR #713, opened 2026-03-12
feat: make python template generator call uv x poetry aware

p2arthur:
📝 Add python_package_manager to Template Context
🚀 Why

We're moving to uv as the default, but our templates don't know that yet! This change passes a variable so the template can decide whether to generate a modern uv project or stay on poetry for older versions.

🛠️ What Changed

Context Injection: Added python_package_manager to the answers_dict we send to Copier.

Smart Default: It defaults to uv if nothing else is set.

Test to check if the worker passes the new  python_package_manager to the provider

The "Handshake": Now templates can just use {% if python_package_manager == 'uv' %} to switch their logic.

### DIFF
--- a/src/algokit/cli/init/command.py
+++ b/src/algokit/cli/init/command.py
@@ -5,6 +5,7 @@ from collections.abc import Callable
 from enum import Enum
 from pathlib import Path
 from typing import NoReturn
+from algokit.core.config_commands.py_package_manager import PyPackageManager, get_py_package_manager
 
 import click
 import prompt_toolkit.document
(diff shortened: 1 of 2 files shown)


aorumbayev (changes requested):
I believe this PR can be closed in favour of a template-side only change in https://github.com/algorandfoundation/algokit-python-template/pull/112. As per discussion with @lempira - the decision was to simply migrate templates to uv instead of poetry, the changes in https://github.com/algorandfoundation/algokit-cli/pull/712 set algokit cli to assume uv by default with minor fallbacks to poetry for backwards compatibility, the templates though should be migrate to uv to avoid complexity of maintaining multiple package managers (which in case of python are less trivial than npm vs pnpm packages that cli supports), plus uv generally faster than poetry. 

Rather than passing python_package_manager from the CLI and having templates conditionally generate poetry vs uv projects, the template (in the pr i opened on python template) now targets uv-only and bumps min_version to v2.11.0. This means:
- No new CLI changes needed — the existing decoupling branch already uses uv for bootstrap/init
- Older CLI versions are blocked at bootstrap time via the min_version check, so there's no risk of a poetry-era CLI hitting a uv-only template
- The template is simpler with no conditional branching between two package managers
The python_package_manager context variable from this PR isn't needed since the template no longer supports poetry at all.

p2arthur:
It makes sense! Supporting UV and poetry automatically would be a bit too much of an edge case. The preferred path should be that our user simply update their algokit cli version to work with latest template versions. Current decoupling branch already implements warnings on that regard, so should be good to go. 

Closing this PR

